PR #2, opened 2024-02-28
Bloch sphere implementation

niccololaurora:
I added a visualization method which implements a Bloch Sphere. This tool could be useful to plot the final states of a classification task.
Example:

`
state = np.array([1, 0], dtype="complex") 
`
`
b_sphere = Bloch()
`
`
b_sphere.add_vector(state, mode="vector", color="black")
`
`
b_sphere.save_plot("b_sphere.pdf")
`
`
b_sphere.plot()
`


src/qiboml/visualization/visualization.py
```diff
@@ -0,0 +1,171 @@
+from mpl_toolkits.mplot3d import Axes3D
+import matplotlib.pyplot as plt
+import numpy as np
+from qibo import hamiltonians
+from qibo.symbols import X, Y, Z
+
+
+class Bloch():
+    def __init__(self):
```
MatteoRobbiati, line 9: I think it would be more natural to use it as follows:

```python
state = circuit.execute().state()
bs = Bloch(state=state)
bs.plot(save=True) 
```

Then one can just update the cached state as `bs.set_state(state=new_state)`.
I think this would be lighter and intuitive.

src/qiboml/visualization/visualization.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def plot(self):
+        "Function to plot the sphere."
+        self.ax.set_aspect("equal")
+        plt.axis('off')
+        plt.tight_layout()
+        plt.show()
+
+    
+    def label_states(self, nclasses):
```
MatteoRobbiati, line 28: This function should be very problem dependent, right?
If your idea is to encode a 3d polygon inside the sphere, then I would suggest to just define a method `bs.distribute_labels_on_surface` or `bs.inscribed_solid(n_vertexes=10)`. This will be very useful to show the inscribed solid and also, if needed, to return the label states corresponding to the vertexes.

MatteoRobbiati (commented):
Thanks a lot for this @niccololaurora!
Some suggestions follow.

src/qiboml/visualization/visualization.py
niccololaurora: Thank you Matteo for the suggestions.

I may be mistaken, but I don't think the concept of a "cached state" would be useful in this context. Based on my experience, I often find myself needing to plot multiple states. Therefore, a natural approach is to create a Bloch object and then add states to it.

```python
state = np.array([1, 0])
bs = Bloch()
bs.add_state(state)
``` 

I will implement the option to pass both a single state (already possible) and a list of states to the add_state function.
Additionally, I have introduced the options of "color" and "mode":

- Color:  In classification tasks, it may be necessary to plot final states associated with different classes, with each class represented by a different color.
- Mode: "vector" or "point". When plotting many states, a point on the sphere is often clearer and cleaner than a cluster of arrows.

Here I have an example:
[sfera_a_q1_l1.pdf](https://github.com/qibogang/qiboml/files/14536843/sfera_a_q1_l1.pdf)]

Finally, I intend to implement a method for blurring the color of points on the sphere based on their distance from the observer's perspective. This would address the challenge of distinguishing points at the front and back of the sphere.
Another possible approach could be to save the image as an .html file, allowing users to interact with the sphere by rotating it to view it from different angles.
I could even implement both things.

niccololaurora:
@MatteoRobbiati we should close this one as well.

PR #4, opened 2024-03-05
Add optimizers to `qiboml`

MatteoRobbiati:
Since in Qibo the optimizers refactor is very slow (due to many older applications and examples based on the old interface), I would suggest to move the optimizers here, so that we can start integrating the automatic differentiation into the optimizers. 

The first two steps I am figuring out are:
1. integrating automatic diff with optimizers;
2. add a `compile` method to the models, which set `optimizer` and `loss_function`, and then write a `model.fit` method on top of the optimizers structure.


src/qiboml/test_reuploading.py
alecandido, line 1: I know this is not yet a Pytest suite, but it could be a starting point. I'd suggest you to immediately place it in `tests/` top-level.

src/qiboml/optimizers/heuristics.py
```diff
@@ -0,0 +1,195 @@
...
+            options=options,
+        )
+
+        return r[1].result.fbest, r[1].result.xbest, r
+
+
+@dataclass
+class BasinHopping(Optimizer):
+    """
+    Global optimizer based on: https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.basinhopping.html.
```
alecandido, line 96: Other minor remark: for this kind of references, you could use intersphinx. I.e. just use the appropriate cross-reference:
```suggestion
    Global optimizer based on: :func:`scipy.optimize.basinhopping`.
```
and then we will configure Sphinx to link SciPy docs as well.

src/qiboml/optimizers/heuristics.py
```diff
@@ -0,0 +1,195 @@
...
+    Note that the Basin-Hopping optimizer combines a global stepping algorithm
```
alecandido, line 97: And basin-hopping are not people names, but a description of action :P
```suggestion
    Note that the basin-hopping optimizer combines a global stepping algorithm
```

PR #7, opened 2024-03-20
Workflows

Edoardo-Pedicillo:
This PR deploy the workflows


Edoardo-Pedicillo:
This PR is ready, but the linting is failing. @MatteoRobbiati could you please fix it?

Edoardo-Pedicillo:
I removed the `pdbpp` dependency, because it is not strictly related to tests  and it cause the following error
https://github.com/qibogang/qiboml/actions/runs/8360680848/job/22887073306#step:11:49, it seems one of its dependency `py3k_compat` is not updated with py 3.10 changes related to `Callable` (see https://stackoverflow.com/questions/69515086/error-attributeerror-collections-has-no-attribute-callable-using-beautifu)

Edoardo-Pedicillo:
After speaking with @alecandido, we agree it is better to take the `pdbpp` dependency and set the dependencies of the `dev` group optional.

MatteoRobbiati:
> After speaking with @alecandido, we agree it is better to take the `pdbpp` dependency and set the dependencies of the `dev` group optional.

Fine with me.

Edoardo-Pedicillo:
@MatteoRobbiati this PR is ready, let me know how many approvals I need.

PR #9, opened 2024-03-28
Divide differentiation and expectation modules

MatteoRobbiati:


src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,91 @@
...
+            ).expectation_from_samples(observable)
+            return expval, grad
+
+        return _expectation_with_tf(params)
+
+    else:
+        raise_error(
+            NotImplementedError,
+            "Only tensorflow automatic differentiation is supported at this moment.",
+        )
```
alecandido, line 91:
```suggestion
    raise_error(
        NotImplementedError,
        "Only tensorflow automatic differentiation is supported at this moment.",
    )
```

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,91 @@
...
+
+    params = circuit.get_parameters()
+    nparams = len(params)
+
+    # read the frontend user choice
+    frontend = observable.backend
+    # construct differentiation backend
+    exec_backend = construct_backend(backend)
+
+    if "tensorflow" in frontend.name:
```
alecandido, line 57: I'd check with `isinstance()` (or otherwise make an `Enum`).

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,91 @@
...
+from typing import List, Optional, Union
+
+import qibo
+from qibo.backends import construct_backend
+from qibo.config import raise_error
+
+from qiboml.operations.differentiation import parameter_shift
+
+
+def expectation_on_backend(
```
alecandido, line 12:
```suggestion
def evaluate(
```
?

Or otherwise just `expectation`.

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,91 @@
...
+                        )
+                    )
+                return gradients
+
+            expval = exec_backend.execute_circuit(
+                circuit=circuit, initial_state=initial_state, nshots=nshots
+            ).expectation_from_samples(observable)
+            return expval, grad
+
+        return _expectation_with_tf(params)
```
alecandido, line 85: Eventually, I'd move it to its own function, and keep the switch-like construct just for the functions' calls.

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if isinstance(frontend, TensorflowBackend):
+        return expectation_with_tf(
+            observable=observable,
+            circuit=circuit,
+            initial_state=initial_state,
+            nshots=nshots,
+            backend=backend,
+            differentiation_rule=differentiation_rule,
+        )
```
alecandido, line 61: Since you will have to repeat this call for all the frontends, maybe it's worth to collect the arguments once (ok, not yet that useful, but the idea is having more than one frontend)

```suggestion
    kwargs = dict(
            observable=observable,
            circuit=circuit,
            initial_state=initial_state,
            nshots=nshots,
            backend=backend,
            differentiation_rule=differentiation_rule,
        )
    if isinstance(frontend, TensorflowBackend):
        return expectation_with_tf(**kwargs)
```

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def expectation_with_tf(
+    observable: qibo.hamiltonians.Hamiltonian,
+    circuit: qibo.Circuit,
+    initial_state: Optional[Union[List, qibo.Circuit]] = None,
+    nshots: int = 1000,
+    backend: str = "qibojit",
+    differentiation_rule: Optional[callable] = parameter_shift,
+):
```
alecandido, line 76: Since you already have the defaults in `expectation`, and that's the user facing one, I'd keep this one defaultless (to have a single source of truth).
```suggestion
def expectation_with_tf(
    observable: qibo.hamiltonians.Hamiltonian,
    circuit: qibo.Circuit,
    initial_state: Optional[Union[List, qibo.Circuit]],
    nshots: int,
    backend: str,
    differentiation_rule: Optional[callable],
):
```

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,111 @@
...
+            values.
+        differentiation_rule (Optional[callable]): the chosen differentiation
+            rule. It can be selected among the methods implemented in
+            ``qiboml.differentiation``.
+    """
+
+    # read the frontend user choice
+    frontend = observable.backend
+
+    if isinstance(frontend, TensorflowBackend):
```
alecandido, line 53: Eventually, we may want to decouple the definition frontend from the execution backend. Not sure how to exactly, but we could consider that.

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,111 @@
...
+            differentiation_rule=differentiation_rule,
+        )
+
+    raise_error(
+        NotImplementedError,
+        "Only tensorflow automatic differentiation is supported at this moment.",
+    )
+
+
+def expectation_with_tf(
```
alecandido, line 69: I believe this should be an internal function. So, you could make explicit that is not meant to be directly used by the user (you can always get the very same result passing through `expectation`, such that there is only one way to do things - less confusing, and easier for us to debug).
```suggestion
def _expectation_with_tf(
```

alecandido: Or even just
```suggestion
def _tf(
```
or
```suggestion
def _with_tf(
```
since you're inside the `expectation` module.

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,111 @@
...
+    Compute expectation sample integrating the custom differentiation rule with
+    TensorFlow's automatic differentiation.
+    """
+    params = circuit.get_parameters()
+    nparams = len(params)
+
+    exec_backend = construct_backend(backend)
+
+    @tf.custom_gradient
+    def _expectation_with_tf(params):
```
alecandido, line 87:
```suggestion
    def _expectation(params):
```
Already inside the `expectation_with_tf` function, thus a very geeneric name should be alright.

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,111 @@
...
+                        initial_state=initial_state,
+                        nshots=nshots,
+                        backend=backend,
+                    )
+                )
+            return gradients
+
+        expval = exec_backend.execute_circuit(
+            circuit=circuit, initial_state=initial_state, nshots=nshots
+        ).expectation_from_samples(observable)
```
alecandido, line 108: Maybe you can even specify if you want the expectation from samples, or the analytic average. Since this is just Python code, it should be alright (you can use `getattr()` on the circuit).

Not necessarily for this PR.

alecandido (commented):
Just some comments for refinements, but it is pretty much what we needed!

Thanks @MatteoRobbiati :D

src/qiboml/operations/expectation.py
alecandido: Yes, but the frontend is derived by the `.backend` of the observable, and the specific frontend you're checking for is called `TensorflowBackend` - which doesn't look much like a frontend :P

From the point of view of Qiboml is no big deal, that's as it is. It is more of a project for Qibo/qibo-core.

MatteoRobbiati: Do you think we can keep this for a next PR?

alecandido: For sure. Even for another repo, potentially.

src/qiboml/operations/expectation.py
```diff
@@ -0,0 +1,138 @@
...
+        return expval, grad
+
+    if differentiation_rule is not None:
+        return _expectation(params)
+
+    elif nshots is None:
+        return _exact(observable, circuit, initial_state, exec_backend)
+
+    else:
+        return _with_shots(observable, circuit, initial_state, nshots, exec_backend)
```
alecandido, line 138: Since they are  not specific to TensorFlow, you could just return the `_expectation` closure and move these into the general `expectation()`

MatteoRobbiati: Done, thanks! :)

alecandido (approved):

PR #14, opened 2024-05-07
Fix and test Jax automatic differentiation

MatteoRobbiati:


src/qiboml/operations/expectation.py
```diff
@@ -44,8 +46,9 @@ def expectation(
             ``qiboml.differentiation``.
     """
 
-    # read the frontend user choice
-    frontend = observable.backend
+    # read the qibo_backend user choice
+    qibo_backend = observable.backend
+    exec_backend = construct_backend(backend)
```
BrunoLiegiBastonLiegi, line 51: I wouldn't construct the backend everytime, maybe it's better to directly ask for the backend object as input

src/qiboml/operations/expectation.py
```diff
@@ -55,8 +58,18 @@ def expectation(
...
     )
-    if isinstance(frontend, TensorflowBackend):
-        return _with_tf(**kwargs)
+
+    if differentiation_rule is not None:
+        if isinstance(qibo_backend, TensorflowBackend):
+            return _with_tf(**kwargs)
+
+        if isinstance(qibo_backend, JaxBackend):
+            return _with_jax(**kwargs)
```
BrunoLiegiBastonLiegi, line 67: maybe the `_with_x` functions should directly be part of the backend, this way here you don't need to check which backend is used but just do `qibo_backend._with(...)`

src/qiboml/operations/expectation.py
```diff
@@ -128,11 +141,53 @@ def grad(upstream):
...
+                gradients.append(
+                    differentiation_rule(
+                        circuit=circuit,
+                        hamiltonian=observable,
+                        parameter_index=p,
+                        initial_state=initial_state,
+                        nshots=nshots,
+                        backend=backend,
+                    )
+                )
```
BrunoLiegiBastonLiegi, line 180: Batched execution wold be great here, for example, as you could perform psr on several different parameters parallely. Namely the batch would be 
```
((theta_0 + eps, theta_1, ..., theta_n),
(theta_0, theta_1 + eps, ..., theta_n),
...,
(theta_0, theta_1, ..., theta_n + eps))
```

src/qiboml/operations/expectation.py
```diff
@@ -128,11 +141,53 @@ def grad(upstream):
...
+    import jax  # pylint: disable=import-error
+
+    params = circuit.get_parameters()
+    exec_backend = construct_backend(backend)
+
+    @jax.custom_gradient
+    def _expectation(params):
+        params = jax.numpy.array(params)
+
+        def grad(params):
```
BrunoLiegiBastonLiegi, line 168: How about jitting grad? Is it possible?

src/qiboml/operations/expectation.py
alecandido: To be fair, it would be even nicer if I could construct a Qibo observable without any specific backend...

src/qiboml/operations/expectation.py
alecandido: It is, but in principle it could be done even later on: JAX transformations are composable, so you could do `grad(grad(jit(grad(...))))`.

src/qiboml/operations/expectation.py
MatteoRobbiati: We need a way to detect which is the frontend, and then build the differentiation rule on top of this. To have an agnostic observable doesn't seems to me a good idea, because it is the way we detect the frontend from the user code. 
Regarding the `_with(...)` suggestion, it could be indeed a nice idea, but then I see a problem: this function will be a third version of the `expectation` we have in the backends. Moreover,  it is a strange expectation since it wants as an argument a `differentiation_rule`, which (my opinion) makes sense only in this QML context.

src/qiboml/operations/expectation.py
MatteoRobbiati: I `jitted` all the functions we are returning now, but I think we should build a Jax version of the Qibojit backend, with a more fine-grained jitting, to boost the performances.

src/qiboml/backends/jax.py
```diff
@@ -1,25 +1,50 @@
...
+        self.np = jnp
+
+    def _cast(self, x, dtype):
+        return self.np.array(x, dtype=dtype)
+
+    def Unitary(self, u):
+        return self._cast(u, dtype=self.dtype)
+
+
 class JaxBackend(NumpyBackend):
```
BrunoLiegiBastonLiegi, line 24: I would probably rewrite the `execute_circuit` and `apply_gate` functions of the backend to properly make use of the jit decorator

BrunoLiegiBastonLiegi (commented):
I am not entirely happy about the structure yet, but I think we need to start implementing our `model` object in order to have a better global view of the thing as a whole, and truly understand which direction to take

PR #20, opened 2024-05-14
Implements the structure of a general `qiboml` model

BrunoLiegiBastonLiegi:
This PR is drafting a possible structure for the `qiboml` models. Namely an abstract `QuantumCircuitLayer` is implemented, which all the other models are going to subclass. 
Three interfaces are going to be provided for this `qiboml` model, namely a `pytorch`, `tensorflow` and `jax` interface.


src/qiboml/models/torch.py~
alecandido, line 1: Gitignore :P

src/qiboml/models/abstract.py
```diff
@@ -0,0 +1,67 @@
...
+    @abstractmethod
+    def backward(self, input_grad: "ndarray"):
+        pass
+
+    @property
+    def parameters(self) -> "ndarray":
+        return self.circuit.get_parameters()
+
+    @parameters.setter
+    def parameters(self, x: "ndarray") -> None:
```
alecandido, line 40: Unfortunately, this is currently a ragged list of tuples (i.e. the elements are gates' parameters), definitely not an array.

I'm trying to normalize this as part of https://github.com/qiboteam/qibo-core/issues/22, but until the refactor, I believe you should declare it as `list`.
```suggestion
    def parameters(self, x: list) -> None:
```

src/qiboml/models/abstract.py
```diff
@@ -0,0 +1,67 @@
...
+
+    nqubits: int
+    qubits: list[int] = None
+    circuit: Circuit = None
+    initial_state: "ndarray" = None
+    backend: "qibo.backends.Backend" = JaxBackend()
+
+    def __post_init__(self) -> None:
+        if self.qubits is None:
+            self.qubits = range(self.nqubits)
```
alecandido, line 24: Not a `list`, so it would not respect the type hint.

Either change the hint, or, simply, just turn this into a `list`:
```suggestion
            self.qubits = list(range(self.nqubits))
```
(range is definitely a more compact representation, but it will only allow for contiguous indices)

src/qiboml/models/abstract.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @parameters.setter
+    def parameters(self, x: "ndarray") -> None:
+        self.circuit.set_parameters(x)
+
+
+@dataclass
+class SequentialQuantumModel(QuantumCircuitLayer):
+
+    layers: list[QuantumCircuitLayer] = None
```
alecandido, line 47: Are you sure we truly need a model for this? Can't we just stack circuits?

(for the time being, this is just a sequence of circuits - that is already a circuit on its own, so I'm not sure there is much benefit in having a dedicated object...)

src/qiboml/models/abstract.py
```diff
@@ -0,0 +1,67 @@
...
+            self.qubits = range(self.nqubits)
+        self.circuit = Circuit(self.nqubits)
+
+    @abstractmethod
+    def forward(self, x):
+        pass
+
+    @abstractmethod
+    def backward(self, input_grad: "ndarray"):
+        pass
```
alecandido, line 33: Why `backward` should differ from `forward`? I can see that in simulation you could have both, but that does not extend to hardware. Is it truly worth to keep both?

src/qiboml/models/ansatze.py
```diff
@@ -1,14 +1,25 @@
...
+        super().__post_init__()
+        for q in self.qubits:
+            self.circuit.add(gates.RY(q, theta=0.0))
+            self.circuit.add(gates.RZ(q, theta=0.0))
+        for i, q in enumerate(self.qubits[:-2]):
+            self.circuit.add(gates.CNOT(q0=q, q1=self.qubits[i + 1]))
+        self.circuit.add(gates.CNOT(q0=self.qubits[-1], q1=self.qubits[0]))
+
+    def forward(self, x: Circuit) -> Circuit:
+        return x + self.circuit
```
alecandido, line 22: This seems closer to function evaluation than forward derivative accumulation...

alecandido (commented):
I haven't dived too much into the encoding part, yet.

src/qiboml/models/abstract.py
```diff
@@ -0,0 +1,67 @@
...
+from qiboml.backends import TensorflowBackend as JaxBackend
+
+
+@dataclass
+class QuantumCircuitLayer(ABC):
+
+    nqubits: int
+    qubits: list[int] = None
+    circuit: Circuit = None
+    initial_state: "ndarray" = None
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Yeah, no, this was actually just a place holder to indicate a general `ndarray` type, either from `numpy`, `torch`, `tf` or `jax`.

src/qiboml/models/abstract.py
BrunoLiegiBastonLiegi: No, this was something that I did in the beginning, I am keeping it just for now, but once the ML interfaces are ready, this will be completely replaced by their own version of `Sequential`.

src/qiboml/models/abstract.py
BrunoLiegiBastonLiegi: The forward and backward pass in ML are two distinct steps. In the forward you feed the data and process them layer by layer until you compute the output of the network, this would be the execution of the circuit in our case. Whereas, the backward pass involves the gradient calculation and aggregation through the chain rule, starting from the output layer and going backward till the input layer, backpropagation indeed. I would rather keep them separated, for the hardware execution case the backward will simply involve several calls of the forward with different parameters.

src/qiboml/models/ansatze.py
BrunoLiegiBastonLiegi: I am not sure to get what you mean, but, indeed, the role of the forward is to process the input data, the backward takes care of evaluating the gradients. Note, anyway, that in most cases, here, the forward is not doing much other than appending a new part to the circuit. The only layers that actually execute the circuit are the decoding layers.

src/qiboml/models/abstract.py
alecandido: This is not always the case, as backward gradient accumulation is only one possible way of computing the gradient, but forward mode is also possible, and convenient according to the problem.
https://en.wikipedia.org/wiki/Automatic_differentiation#Forward_and_reverse_accumulation
(the trade-off is between the number of inputs and outputs, i.e. the Jacobian shape)

If you want to keep function evaluation separate from the gradient one is perfectly fine, but I'd choose at least different names from *forward* and *backward*, and what I was saying is that most likely you will need a single name for the gradient (i.e. we don't want to implement evaluation + backward-gradient + forward-gradient, but just evaluation + gradient).

src/qiboml/models/ansatze.py
alecandido: I believe the misunderstanding here is the same of the previous comment. Explanation in https://github.com/qiboteam/qiboml/pull/20#discussion_r1627318244

BrunoLiegiBastonLiegi:
@alecandido @MatteoRobbiati  This is more or less ready for a first complete review. The `torch` and `keras` interfaces are implemented and 'tested' (just a forward pass) in the notebook inside `tutorials/`. The `torch` interface seems to work fine but we have a major problem with `tf` and `keras`: 
``` python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[1], line 19
     17 for x in data:
     18     x = tf.expand_dims(x, axis=0)
---> 19     print(model(x).probabilities())

File ~/python_envs/qibo/lib/python3.10/site-packages/keras/src/utils/traceback_utils.py:122, in filter_traceback.<locals>.error_handler(*args, **kwargs)
    119     filtered_tb = _process_traceback_frames(e.__traceback__)
    120     # To get the full stack trace, call:
    121     # `keras.config.disable_traceback_filtering()`
--> 122     raise e.with_traceback(filtered_tb) from None
    123 finally:
    124     del filtered_tb

File ~/python_envs/qibo/lib/python3.10/site-packages/tensorflow/python/framework/constant_op.py:108, in convert_to_eager_tensor(value, ctx, dtype)
    106     dtype = dtypes.as_dtype(dtype).as_datatype_enum
    107 ctx.ensure_initialized()
--> 108 return ops.EagerTensor(value, ctx.device_name, dtype)

ValueError: Exception encountered when calling BinaryEncodingLayer.call().

Attempt to convert a value (<qibo.models.circuit.Circuit object at 0x7b69c41af730>) with an unsupported type (<class 'qibo.models.circuit.Circuit'>) to a Tensor.

Arguments received by BinaryEncodingLayer.call():
  • x=tf.Tensor(shape=(1, 5), dtype=float32)
```

If I am interpreting this correctly, tensorflow tries to cast the output of each layer to an `EagerTensor`, which is clearly not possible for our encoding and ansatz layers that just return a `qibo.Circuit`. Honestly, I am not sure if there is an easy workaround for this, but this seems a major problem as it conflicts with the fundamental design idea I had in mind...

BrunoLiegiBastonLiegi:
@alecandido @MatteoRobbiati I just finished fixing the `keras` interface and writing the initial tests. Now, we are able to expose our quantum layers as a `QuantumModel` that is in facts either a `torch.nn.Module` or a `keras.layers.Layer`. To be more precise, the two `QuantumModel`s are probably going to need some refactoring to support custom differentiation, but for now I think it might be good to have a minimal first version and close this with just the forward pass working. 
Regarding the tests, I added the `backend` fixture (same as in `qibo`), that currently tests `numpy`, `torch` and `tensorflow`, and the `frontend` fixture that tests the `torch` and `keras` interfaces. The tests are rather minimal, in the sense that I am just checking that a forward pass terminates without crashing, I could fix the seed and come up with some final targets, but my plan was to test that a loss function was decreasing after doing a couple of parameter update steps (therefore once the differentiation will be working).
There is a problem with the `windows` test that I am not fully understanding, but the tests with the other platforms seem to be working fine.
I still need to remove some leftovers here and there (like the `backward` method that is probably no longer needed) but I would say that you could start looking at this.

BrunoLiegiBastonLiegi:
@MatteoRobbiati is the `JaxBackend` tested anywhere? Because running a simple circuit fails on mac for me. Probably the compatibility with apple METAL is still quite experimental, it is maybe better to disable GPU acceleration for the moment on `darwin`. What do you think @alecandido?

alecandido:
> @MatteoRobbiati is the `JaxBackend` tested anywhere? Because running a simple circuit fails on mac for me. Probably the compatibility with apple METAL is still quite experimental, it is maybe better to disable GPU acceleration for the moment on `darwin`. What do you think @alecandido?

Certainly, it is not a priority in any case, since heavy GPU jobs would never run on Apple (and for light jobs the CPU is enough).
I'm in favor of making our life easier, and focusing on relevant tasks :)

MatteoRobbiati:
> @MatteoRobbiati is the `JaxBackend` tested anywhere? Because running a simple circuit fails on mac for me. Probably the compatibility with apple METAL is still quite experimental, it is maybe better to disable GPU acceleration for the moment on `darwin`. What do you think @alecandido?

I tested it locally (only CPUs) some time ago when I was implementing it into Qibo. I never tested GPUs indeed.

BrunoLiegiBastonLiegi:
Moreover, `qiboml`'s pytorch is outdated, I am using `qibo`'s one for the moment.

src/qiboml/models/keras.py
```diff
@@ -0,0 +1,65 @@
+"""Keras interface to qiboml layers"""
+
+from dataclasses import dataclass
+
+import keras
```
alecandido, line 5: You're using Keras, but that's undeclared as a dependency in `pyproject.toml`.

src/qiboml/models/pytorch.py
```diff
@@ -0,0 +1,63 @@
...
+                    torch.nn.Parameter(torch.as_tensor(layer.circuit.get_parameters())),
+                )
+        if not isinstance(self.layers[-1], ed.QuantumDecodingLayer):
+            raise_error(
+                RuntimeError,
+                f"The last layer has to be a `QuantumDecodinglayer`, but is {self.layers[-1]}",
+            )
+
+    def forward(self, x: torch.Tensor):
+        if self.backend.name != "pytorch":
```
alecandido, line 44: Any reason to use PyTorch with a non-Pytorch backend?

alecandido (commented):
Preliminary review

src/qiboml/models/abstract.py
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: I could try to set up a workflow with `mypy` then, should I add it to the `qiboteam/workflows` repo and then load it here from there?

src/qiboml/models/abstract.py
```diff
@@ -0,0 +1,40 @@
...
+    backend: "qibo.backends.Backend" = JaxBackend()
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: In this cases I am not sure whether it is worth importing objects just for type hinting. Here it should not be a problem, but I remember that in `qibo` I often encountered circular import errors.

src/qiboml/operations/expectation.py
```diff
@@ -56,7 +56,7 @@ def expectation(
         initial_state=initial_state,
         nshots=nshots,
         differentiation_rule=differentiation_rule,
-        exec_backend=exec_backend,
+        # exec_backend=exec_backend,
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: This comes from the main, but it fails because there is no `exec_backend` argument, I commented this out here because I am not sure how this is supposed to work, maybe @MatteoRobbiati can provide some insight about this.

src/qiboml/models/keras.py
```diff
@@ -0,0 +1,65 @@
...
+import numpy as np
+import tensorflow as tf  # pylint: disable=import-error
+from qibo.config import raise_error
+
+import qiboml.models.encoding_decoding as ed
+from qiboml.models.abstract import QuantumCircuitLayer
+
+
+@dataclass
+class QuantumModel(keras.layers.Layer):  # pylint: disable=no-member
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Let's say that it is more in a ML style, and indeed as you are saying you only need encoding, circuit and decoding, that you could see as three different layers stacked on top of each other. The only difference is that here you have the flexibility to construct also the circuit modularly, seen as different layers.

Regarding the encoding, I am not sure it's fully classical, in the sense that the input is usually classical but the output should be quantum in general. Instead, the decoding could be an observable, yes, but I would like to offer the possibility of returning other things, like for examples the probabilities or the samples.

In any case the `QuantumLayer` you are suggesting is implemented here already, it's just called in a different way. Namely, the `keras` interface `qiboml.models.keras.QuantumModel` with an `ExpectationLayer` as decoding is precisely taking care of that.

src/qiboml/models/keras.py
```diff
@@ -0,0 +1,65 @@
...
+            raise_error(
+                RuntimeError,
+                f"The last layer has to be a `QuantumDecodinglayer`, but is {self.layers[-1]}",
+            )
+
+    def call(self, x: tf.Tensor) -> tf.Tensor:
+        if self.backend.name != "tensorflow":
+            if self.backend.name == "pytorch":
+                self.backend.requires_grad(False)
+            x = self.backend.cast(np.array(x))
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: This is implementing the `keras` interface, meaning that you expect as input and output a `tf.Tensor`, however your backend could be any of the `qibo` backends. For this reason you need casts if you are not using the `TensorflowBackend`.

src/qiboml/models/pytorch.py
BrunoLiegiBastonLiegi: Similarly to the comment above, this is the `pytorch` interface, which consumes `torch.tensors` but this is separated from the backend you are using. If the backend is the `PyTorchBackend`, then you don't need to do anything, otherwise you have to cast to the appropriate type.

src/qiboml/models/abstract.py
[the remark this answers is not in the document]
alecandido: Worfklows could be annoying. Workflows in another repo even more.

I'd suggest going in steps: do it here, and then we move to the `qiboteam/workflows`.

Notice that:
- Qibosoq already has it https://github.com/qiboteam/qibosoq/blob/main/.github/workflows/mypy.yml
- there are already issues for that in `qiboteam/workflows`, I've simply never taken the time of implementing because there would have been no one using it
  - https://github.com/qiboteam/workflows/issues/31
  - https://github.com/qiboteam/workflows/issues/32 (`symilar` is part of Pylint, might be needed only when we'll remove Pylint)

src/qiboml/models/abstract.py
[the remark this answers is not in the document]
alecandido: Yes, the circular imports are a problem. But the strings are not the solution (just a patch).

The main reason for allowing strings are mutual references. Which should be possible (valid at runtime) but not working.
There are many PEPs for postponed annotation, and it's the only feature that is still part of `__future__`, and delayed for 4.0.

In any case, I can give you examples of why the proper solution would be different when we'll encounter them. But it should never be the case for a `qibo` object in `qiboml` (since `qibo` should never import `qiboml`, not *statically*).

src/qiboml/models/keras.py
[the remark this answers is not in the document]
alecandido: I have the feeling you're still using the `keras` inside TensorFlow, and not the separate Keras distribution...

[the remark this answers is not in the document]
alecandido: https://keras.io/api/layers/core_layers/input/

It has no framework (nor content) associated.

src/qiboml/models/keras.py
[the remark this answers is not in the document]
alecandido: > The only difference is that here you have the flexibility to construct also the circuit modularly, seen as different layers.

But this you already have: different circuits are already "layers" that you can concatenate.
Simply, there are red layers and blue layers (quantum and classical) that doesn't make sense to mix. So, to me, this is better implemented if you don't provide (and then possibly deny) the option of concatenating them. But simply keeping them as different and non-compatible objects.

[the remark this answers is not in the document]
alecandido: > Regarding the encoding, I am not sure it's fully classical, in the sense that the input is usually classical but the output should be quantum in general.

In which sense the output is quantum? Usually you encode in parameters.

If instead you're encoding in a state, this is already cutting Qiboml out of hardware support. Much better than to have a function that is emitting a circuit (thus homogeneous to the *circuit* part) rather than a tensor.

My take on the goal of Qiboml is more in the direction of making Qibo compatible with machine learning, not to have an ML-practitioner-friendly interface to Qibo (since Qibo already has an interface, and doesn't need a wrapper).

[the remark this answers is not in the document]
alecandido: > Consider for example the binary or phase encoding, you have some classical inputs and you convert them in a layer of X or rotation gates. Of course the gate are represented by the index of the qubit they act on or the parameter, which are classical data, but intrinsically the encoding takes classical data and embeds it in a quantum circuit. Thus I wouldn't call it fully classical.

I would classify the X as part of the circuit, and the indexing as part of the encoding. You can not make a *previous* layer that is stacked before the circuit, since your encoding might also be in the middle of the circuit. In that case, it doesn't make sense to see the rest of the circuit as a separate layer from the encoding.

src/qiboml/models/pytorch.py
BrunoLiegiBastonLiegi: Yes indeed, but since there is currently no implementation of the `evaluate` function, as gradients are still not in main yet, I had to take care of that here in order to test. As I said in one of the previous meetings, this object is probably going to change when we incorporate the auto differentiation, as pytorch wants you to write a custom autograd function which is probably going to perform what is done here in the `forward`.

alecandido: Ok, maybe it's worth to commit a bit of time to the Pytorch implementation, since (in principle) `expectation()` is in `main`:
https://github.com/qiboteam/qiboml/blob/0e906eef4117da995e3897c7b50c92c8bf6e6a1e/src/qiboml/operations/expectation.py#L12-L19
(sorry, I messed up, it's not `evaluate()`)

Working on top of that, should make the model structure more compatible.

In any case, if you prefer doing it in a different PR I'm not completely against. But you might end up reverting there part of what is done here.

src/qiboml/models/abstract.py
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Ok I created a `qiboml.ndarray` type that groups the array types of all the installed frameworks: numpy (and jax) is always there, whereas `torch.Tensor` and `tf.Tensor` are present only when installed.

src/qiboml/models/abstract.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __post_init__(self) -> None:
+        if self.qubits is None:
+            self.qubits = list(range(self.nqubits))
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: The problem I see here, though, is that then you will need to construct the layer as:
```
QuantumCircuitLayer(nqubits=5, _qubits=[0,1,2])
```
if you want to specify the qubits. I am not sure if there's any workaround with `dataclasses`.

src/qiboml/models/keras.py
BrunoLiegiBastonLiegi: I double-checked this and it seems that `keras` gets installed through the `tensorflow` dependency, do you think it is better to have it explicitely declared anyway?

src/qiboml/models/abstract.py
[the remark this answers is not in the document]
alecandido: Interestingly enough, you can do it with Pydantic:
```py
class Ciao(BaseModel):
    name_: str = Field(alias="name")
                                                                                                                                                                                      
    @property
    def name(self):
        print(f"Ciao, {self.name_}")
        return self.name_
                                                                                                                                                                                      
c = Ciao(name="come va?")
c.name
```
(though you can not use `._qubit`, since fields are supposed to be public, and they can not contain leading underscores)

alecandido (approved):
Just one simple criticality spotted top-level.

About the longer-standing discussions, there is no need to solve in this PR anyhow. Qiboml is an experimental module, and it will keep being experimental even after the PR.

We should take care about having a programmatic discussion before more serious attempts towards stabilization, but it is not the yet the time for this. Keep experimenting is fine, and moving faster is definitely a priority (though collecting observations along the why is good, as well as keep discussing them - and we should still pay attention to more prosaic software aspects, as correct import/dependencies, and of course tests :P).

pyproject.toml
MatteoRobbiati, line 1: I updated the lock in my PSR branch to the purpose of this PR fine with this!

src/qiboml/models/ansatze.py
MatteoRobbiati, line 1: As discussed, this is fine as long it is temporary: we need a combined training-encoding strategy to compute something actually useful. But yeah, we opened issues for this and we know. So Fine as it is for this PR.

MatteoRobbiati (approved):
Let's go ahead with this one. 
I have comments but many of them have been implemented in #42.
I would merge this and fix all those things after.

PR #26, opened 2024-06-17
Moving expval calculation into backends and adding torch differentiation

MatteoRobbiati:


alecandido:
Why does Qiboml need a further function in the backend?

This is coupling once more the frontends with the backends, and defeating the original purpose of making them independent.
So, to have a circuit on hardware (`QibolabBackend`) differentiable by Pytorch, we will need to maintain a whole `PytorchBackend`, same for each possible other library, that in principle could be extended even beyond TensorFlow + Pytorch + JAX at the price of implementing a single function, not a whole backend.

Execution and derivative presentation are completely separate domain, thus we should keep them separate to preserve modularity.

MatteoRobbiati:
> Why does Qiboml need a further function in the backend?
> 
> This is coupling once more the frontends with the backends, and defeating the original purpose of making them independent. So, to have a circuit on hardware (`QibolabBackend`) differentiable by Pytorch, we will need to maintain a whole `PytorchBackend`, same for each possible other library, that in principle could be extended even beyond TensorFlow + Pytorch + JAX at the price of implementing a single function, not a whole backend.
> 
> Execution and derivative presentation are completely separate domain, thus we should keep them separate to preserve modularity.

This is extremely true. Thanks for catching this before further modifications. While experimenting, yesterday, I lost this fundamental point. 
I'll move all the `torch` features into the `sym` backend (that part is working anyway) and close this.

PR #41, opened 2024-10-14
Fixes related to TF migration

MatteoRobbiati:
A couple of modifications here are temporary and needed to make tests working both in Qibo and Qiboml.

### DIFF
--- a/src/qiboml/backends/__init__.py
+++ b/src/qiboml/backends/__init__.py
@@ -6,7 +6,7 @@ from qiboml.backends.jax import JaxBackend
 from qiboml.backends.pytorch import PyTorchBackend
 from qiboml.backends.tensorflow import TensorflowBackend
 
-PLATFORMS = ["tensorflow", "pytorch", "jax"]
+PLATFORMS = ["tensorflow"]
 QibomlBackend = Union[TensorflowBackend, PyTorchBackend, JaxBackend]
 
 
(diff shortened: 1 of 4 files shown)


src/qiboml/backends/__init__.py
```diff
@@ -7,6 +7,7 @@
 from qiboml.backends.tensorflow import TensorflowBackend
 
 PLATFORMS = ["tensorflow", "pytorch", "jax"]
+AVAILABLE_PLATFORMS = ["tensorflow"]
```
BrunoLiegiBastonLiegi, line 10:
```suggestion
AVAILABLE_PLATFORMS = ["tensorflow"] # temporary: to remove once pytorch and tensorflow are migrated  and jax is fully woring
```

this is only temporary, right? I would write a reminder then

BrunoLiegiBastonLiegi (approved):

src/qiboml/backends/__init__.py
```diff
@@ -7,7 +7,7 @@
 from qiboml.backends.tensorflow import TensorflowBackend
 
 PLATFORMS = ["tensorflow", "pytorch", "jax"]
-AVAILABLE_PLATFORMS = ["tensorflow"]
+AVAILABLE_PLATFORMS = ["tensorflow"] # temporary: to remove once pytorch and tensorflow are migrated  and jax is fully woring
```
renatomello, line 10:
```suggestion
AVAILABLE_PLATFORMS = ["tensorflow"] # temporary: to remove once pytorch and tensorflow are migrated  and jax is fully working
```

PR #42, opened 2024-10-22
Enabling PSR, analytical diff with pytorch and moving interfaces to dedicated module

MatteoRobbiati:
Automatic differentiation supported for Pytorch interface and with Torch analytical and PSR methods.


tests/test_differentiation_rules.py
```diff
@@ -0,0 +1,91 @@
...
+
+    expval = q_model(x)
+    expval.backward()
+
+    # TODO: standardize this output with keras' one and use less convolutions
+    grad = interface_engine.to_numpy(list(q_model.parameters())[-1].grad)
+
+    assert np.round(grad[0], decimals=decimals) == np.round(
+        TARGET_GRAD[0], decimals=decimals
+    )
```
BrunoLiegiBastonLiegi, line 88: what about `grad[1]`? not sure exactly which shape `grad` has

src/qiboml/interfaces/pytorch.py
```diff
@@ -25,30 +17,23 @@ class QuantumModel(torch.nn.Module):
...
         params.requires_grad = True
         self.circuit_parameters = torch.nn.Parameter(params)
 
-        if self.differentiation == "auto":
-            self.differentiation = BACKEND_2_DIFFERENTIATION.get(
-                self.backend.name, "PSR"
-            )
-
-        if self.differentiation is not None:
-            self.differentiation = getattr(Diff, self.differentiation)()
```
BrunoLiegiBastonLiegi, line 46: what happens now if the backend is not `pytorch`, say `numba`, and `differentiation_rule=None`? I believe it will crash, because it will try to use the `QuantumModelAutoGrad` with no differentiation rule.

tests/test_models_interfaces.py
```diff
@@ -251,12 +256,19 @@ def test_decoding(backend, frontend, layer, seed, analytic):
...
+            differentiation_rule = PSR()
+        else:
+            differentiation_rule = None
+        kwargs["nshots"] = None
     decoding_layer = layer(nqubits, decoding_qubits, **kwargs)
 
-    if not decoding_layer.analytic:
-        pytest.skip("PSR differentiation is not working yet.")
+    # if not decoding_layer.analytic:
+    #     pytest.skip("PSR differentiation is not working yet.")
```
BrunoLiegiBastonLiegi, line 267:
```suggestion
```

src/qiboml/__init__.py
```diff
@@ -12,7 +12,7 @@
 try:
     from tensorflow import Tensor as tf_tensor
 
-    from qiboml.models import keras
+    from qiboml.interfaces import keras
```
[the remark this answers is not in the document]
MatteoRobbiati: I would personally keep interface, but open to eventual refactor :)

src/qiboml/operations/differentiation.py
```diff
@@ -1,23 +1,42 @@
...
+
+@dataclass
+class DifferentiationRule(ABC):
 
+    @abstractmethod
+    def evaluate(
+        self,
+        x: ndarray,
+        encoding: QuantumEncoding,
+        training: Circuit,  # TODO: replace with an abstract TrainableLayer
```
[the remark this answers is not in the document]
MatteoRobbiati: I added a point to our list #44.

src/qiboml/models/decoding.py
```diff
@@ -42,9 +41,17 @@ def set_backend(self, backend):
     def output_shape(self):
         raise_error(NotImplementedError)
 
+    @property
+    def analytic(self):
+        if self.nshots is None:
+            return True
+        else:
+            return False
```
renatomello, line 49:
```suggestion
        return False
```

src/qiboml/operations/differentiation.py
```diff
@@ -35,18 +35,85 @@ def evaluate(
...
+        circuit.set_parameters(tmp_params)
+        forward = decoding(circuit)
+
+        tmp_params = self.shift_parameter(tmp_params, parameter_index, -2 * s, backend)
+
+        circuit.set_parameters(tmp_params)
+        backward = decoding(circuit)
+        return generator_eigenval * (forward - backward)
+
+    def gradient_wrt_data(
```
renatomello, line 95:
```suggestion
    def gradient_from_data(
```

src/qiboml/operations/differentiation.py
```diff
@@ -1,42 +1,163 @@
...
+                )
+            )
+        return gradient
+
+    def one_parameter_shift(
+        self, circuit, decoding, parameters, parameter_index, backend
+    ):
+        """Compute one derivative of the decoding strategy w.r.t. a target parameter."""
+        gate = circuit.associate_gates_with_parameters()[parameter_index]
+        generator_eigenval = gate.generator_eigenvalue()
```
BrunoLiegiBastonLiegi, line 80: One question, this has to be computed every time, right? There is no way of caching it, because it depends on the value of the parameter, right?

src/qiboml/operations/differentiation.py
```diff
@@ -1,42 +1,163 @@
...
+        circuit.set_parameters(tmp_params)
+        forward = decoding(circuit)
+
+        tmp_params = self.shift_parameter(tmp_params, parameter_index, -2 * s, backend)
+
+        circuit.set_parameters(tmp_params)
+        backward = decoding(circuit)
+        return generator_eigenval * (forward - backward)
+
+    def gradient_from_data(
```
BrunoLiegiBastonLiegi, line 95: So, i am not supposed to look into this, right?

src/qiboml/operations/differentiation.py
```diff
@@ -105,155 +226,11 @@ def _run_without_inputs(self, *parameters):
...
-
-    return result
+def shifted_x_component(
+    x: ndarray, index: int, shift_value: float, backend: Backend
+) -> float:
+    """Shift a component of an ndarray."""
+    flat_array = backend.to_numpy(x).flatten()
+    shifted_flat_array = deepcopy(flat_array)
+    shifted_flat_array[index] += shift_value
+    return shifted_flat_array[index]
```
BrunoLiegiBastonLiegi, line 236: what's this for?

tests/test_differentiation_rules.py
```diff
@@ -0,0 +1,100 @@
...
+    parameters. In this test the system size is fixed to two qubits and all the
+    parameters/data values are fixed.
+    """
+
+    if frontend.__name__ == "qiboml.interfaces.keras":
+        pytest.skip("keras interface not ready.")
+    elif frontend.__name__ == "qiboml.interfaces.pytorch":
+        # TODO: replace with qiboml, pytorch as soon as migration is complete
+        # TODO: define a proper qiboml.set_interface() procedure for these situations
+        qibo.set_backend("pytorch")
```
BrunoLiegiBastonLiegi, line 59: This I don't think is needed, as you have explicitely defined the `EXECUTION_BACKENDS`

tests/test_differentiation_rules.py
```diff
@@ -0,0 +1,100 @@
...
+    decoding_layer = Expectation(
+        nqubits=nqubits,
+        backend=backend,
+        observable=obs,
+        nshots=nshots,
+    )
+
+    nparams = len(training_layer.get_parameters())
+    initial_params = np.linspace(0.0, 2 * np.pi, nparams)
+    training_layer.set_parameters(initial_params)
```
BrunoLiegiBastonLiegi, line 84:
```suggestion
    training_layer.set_parameters(backend.cast(initial_params))
```
just for the sake of consistency, it should not make any difference in practice

tests/test_models_interfaces.py
```diff
@@ -220,14 +221,17 @@ def test_encoding(backend, frontend, layer, seed):
...
-def test_decoding(backend, frontend, layer, seed, analytic):
-    if frontend.__name__ == "qiboml.models.keras":
+def test_decoding(backend, frontend, layer, seed):
+    if frontend.__name__ == "qiboml.interfaces.keras":
         pytest.skip("keras interface not ready.")
     if backend.name not in ("pytorch", "jax"):
         pytest.skip("Non pytorch/jax differentiation is not working yet.")
-    if analytic and not layer is dec.Expectation:
+    if layer.analytic and not layer is dec.Expectation:
         pytest.skip("Unused analytic argument.")
```
BrunoLiegiBastonLiegi, line 230:
```suggestion
```

this is not needed anymore I believe, as `analytic` is not a pytest parameter anymore

BrunoLiegiBastonLiegi (commented):
Is the notebook working? 
In any case I would put it in a separate PR as it is outside of the scope of this PR.

pyproject.toml
```diff
@@ -14,10 +14,10 @@ packages = [{ include = "qiboml", from = "src" }]
 python = ">=3.9,<3.13"
 numpy = "^1.26.4"
 keras = { version = "^3.0.0", optional = true }
-tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux' or sys_platform == 'darwin'", optional = true }
+tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux' or sys_platform == 'darwin'"}
```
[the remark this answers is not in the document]
MatteoRobbiati: Reminder: add it to test deps

src/qiboml/models/encoding.py
```diff
@@ -26,12 +26,19 @@ def __post_init__(
...
+        pass
+
     @property
     def circuit(
         self,
     ):
         return self._circuit
 
+    @property
+    def hardware_differentiable(self):
```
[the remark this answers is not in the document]
MatteoRobbiati: I will remove this since it is not used! It was supposed to be used in the old PSR wrt data version.

src/qiboml/operations/differentiation.py
MatteoRobbiati: In the end this value is returned as an attribute of the specific gate, I think, as discussed, we can keep it as it is and in case re-discuss the PSR tools in Qibo.

src/qiboml/operations/differentiation.py
MatteoRobbiati: It was used for grad wrt inputs. Now removed since that grad contribution is padded with zeros (for now).

src/qiboml/interfaces/pytorch.py
```diff
@@ -25,29 +18,22 @@ class QuantumModel(torch.nn.Module):
...
     decoding: QuantumDecoding
-    differentiation: str = "auto"
+    differentiation: Differentiation = None
 
     def __post_init__(
         self,
     ):
         super().__init__()
 
+        circuit = self.encoding.circuit
```
BrunoLiegiBastonLiegi, line 28: what is this for? I don't think it is used anywhere below.
```suggestion
```

src/qiboml/interfaces/pytorch.py
BrunoLiegiBastonLiegi: @MatteoRobbiati  I think we still have some problems with the automatic choice of the proper differentiation method. For instance, if you were not using the pytorch backend or your decoding was not analytically differentiable and you don't specify the `differentiation` (which is `None` by default) the `forward` will crash.

BrunoLiegiBastonLiegi (commented):
Thanks for the update, the only thing that seems still up is this https://github.com/qiboteam/qiboml/pull/42#discussion_r1905740704. For the rest, once we extend the coverage the best we can this looks ready to be merged. 

Btw, concerning the notebook included here, is it ready to be merged as well or is it a draft? Because I did not look into it.

src/qiboml/interfaces/pytorch.py
```diff
@@ -25,41 +19,35 @@ class QuantumModel(torch.nn.Module):
...
     def __post_init__(
         self,
     ):
         super().__init__()
 
         params = [p for param in self.circuit.get_parameters() for p in param]
-        params = torch.as_tensor(self.backend.to_numpy(params)).ravel()
+        params = torch.as_tensor(self.backend.to_numpy(x=params)).ravel()
         params.requires_grad = True
         self.circuit_parameters = torch.nn.Parameter(params)
```
BrunoLiegiBastonLiegi, line 33: restore the `BKD2DIFF` global dict with default choices:
pytorch -> none
jax -> jax
tf -> jax
other -> PSR

BrunoLiegiBastonLiegi, line 33: check if decoding.analytic, if not -> PSR, otherwise the `BK2DIFF` mechanism

tutorials/model_training.ipynb
BrunoLiegiBastonLiegi, line 1: remove this

BrunoLiegiBastonLiegi (approved):
Thanks, I think this can go to main now. Just double check whether you want to keep the notebook.

edit: ah ok I did not notice there was a test failing, let me have a look.

PR #45, opened 2024-10-30
Temporary fixes to support pytorch backend migration

MatteoRobbiati:
This is needed to complete the Pytorch backend migration from Qibo to Qiboml.


BrunoLiegiBastonLiegi:
Are these just the recent modifications added in qiboteam/qibo#1450 ?

src/qiboml/backends/__init__.py
```diff
@@ -7,7 +7,10 @@
 from qiboml.backends.tensorflow import TensorflowBackend
 
 PLATFORMS = ["tensorflow", "pytorch", "jax"]
-AVAILABLE_PLATFORMS = ["tensorflow"] # temporary: to remove once pytorch and tensorflow are migrated  and jax is fully working
+AVAILABLE_PLATFORMS = [
+    "tensorflow",
+    "pytorch",
+]  # temporary: to remove once pytorch and tensorflow are migrated  and jax is fully working
```
BrunoLiegiBastonLiegi, line 13: what's the plan here? now that we are migrating pytorch, we could get rid of this, right?

BrunoLiegiBastonLiegi (approved):

renatomello (approved):

src/qiboml/backends/__init__.py
MatteoRobbiati: I would keep it until Jax is ready, if we still plan to support pytorch soon.

PR #48, opened 2024-11-04
Introducing trainable-encoding layers

MatteoRobbiati:


src/qiboml/models/ansatze.py
```diff
@@ -17,14 +17,17 @@ def layered_ansatz(
     nqubits: int,
     nlayers: int = 1,
     qubits: list[int] = None,
-    gates_list: Optional[List[gates.Gate]] = [gates.RY, gates.RZ],
+    gates_list: Optional[List[gates.Gate]] = [
+        gates.RY,
+        gates.RZ,
+    ],  # TODO: this has to be a circuit
```
renatomello, line 26: Correct me if I'm wrong, but it seems like this entire file is a duplication of `qibo.models.encodings.phase_encoder` and `qibo.models.encodings.entangling_layer`.

MatteoRobbiati: Yes that is true! Those are just small and simple functions to make qiboml work now :)

But they are definitely temporary

Issue #47, opened 2024-11-01
 Symbolic Tensors 


niccololaurora:
The Keras interface works fine but I had to change the Binary Encoding.

TensorFlow’s Graph Mode relies on Symbolic Tensors, which act as placeholders without actual data during initial model checks. When using ```model.fit```, TensorFlow first uses these Symbolic Tensors to traverse and verify the computation graph, ensuring compatibility across all operations. Only after this validation process, when the graph structure is confirmed as sound, does TensorFlow pass real data through the model to perform actual training. This approach allows TensorFlow to optimize and catch issues before handling real inputs.

Therefore, we have the following problems (probably more) when dealing with Symbolic Tensors:

1. We cannot use NumPy functions on Symbolic Tensors;
2. We cannot access data contained in Symbolic Tensors (because they are empty);
3. We cannot iterate over Symbolic Tensors (because they are empty);
4. We cannot use "if" statements if Symbolic Tensors are involved;

We need to work around these problems by using native TensorFlow functions.
For instance, instead of an "if" statement we must use ```tf.cond```.

However, the issue that I have encountered with the Binary Encoding (which has forced me to redefine this encoding) could be a sign of a deeper problem: we might need to define each encoding (or every other component of the QuantumModel) depending on the frontend that we use.
This is due to the fact that Symbolic Tensors force us to use native TensorFlow functions, that could be not compatible with PyTorch.

BrunoLiegiBastonLiegi:
You can force eager execution with `run_eagerly=True` in `model.compile` [https://www.tensorflow.org/api_docs/python/tf/keras/Model#run_eagerly](https://www.tensorflow.org/api_docs/python/tf/keras/Model#run_eagerly)
This is probably not recommended for the sake of pure speed, but we could just do that for those layers that are problematic, at least for the moment.

BrunoLiegiBastonLiegi:
#54

PR #54, opened 2024-11-20
Keras Interface 

niccololaurora:
This is the implementation of the Keras Interface. This PR supports only the differentiation with TensorFlow as backend.
In a future PR, I plan to enhance this Keras Interface by adding support for other types of differentiation, such as JAX.

As stated in issue #47, I had to change the Binary encoding implementation to use Symbolic Tensors.


.gitignore
```diff
@@ -127,6 +127,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+qiboml/
```
BrunoLiegiBastonLiegi, line 130:
```suggestion
```
better not to ignore this

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
...
+            "justMyCode": false,
+            "console": "integratedTerminal",
+            "env": {
+                "ON_HEROKU": "0",
+                "PYTEST_ADDOPTS": "-c pytest.ini",
+                "ECHO_SQL_QUERIES": "1"
+            },
+        }
+    ]
+}
```
BrunoLiegiBastonLiegi, line 21: this is probably some vscode leftover, better to remove it

prova.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def circuit(self):
+        c = Circuit(2)
+        c.add(gates.X(0))
+        c.add(gates.RX(1, theta=0.5))
+
+    def call(self, x):
+        self.circuit()
+
+    def
```
BrunoLiegiBastonLiegi, line 35: this has to be removed as well

pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+env = 
+    TESTING=true
+    ENV=local
```
BrunoLiegiBastonLiegi, line 4: even this has to go I would say

src/qiboml/models/encoding.py
```diff
@@ -44,8 +46,9 @@ def __post_init__(
             self._circuit.add(gates.RY(q, theta=0.0, trainable=False))
 
     def _set_phases(self, x: ndarray):
-        for gate, phase in zip(self._circuit.parametrized_gates, x.ravel()):
-            gate.parameters = phase
+        phase = tf.reshape(x, [-1])
```
BrunoLiegiBastonLiegi, line 49: we cannot have `tf` functions here, this has to work for `pytorch` as well and eventually `jax` when we introduce the `Flax` interface

src/qiboml/models/encoding.py
```diff
@@ -62,7 +65,15 @@ def __call__(self, x: ndarray) -> Circuit:
...
+        def true_fn():
+            circuit.add(gates.X(q))
+
+        def false_fn():
+            tf.no_op()
+
+        for i, q in enumerate(self.qubits):
+            pred = tf.equal(x[0][i], 1)
+            tf.cond(pred, true_fn=true_fn, false_fn=false_fn)
+
```
BrunoLiegiBastonLiegi, line 78: Same here, the separation of true and false branches is fine, but you can't use `tf` functions. In any case, #55 is slightly changing the binary encoding and should lift the problems related to `tf` symbolic tensors here

src/qiboml/models/keras.py
```diff
@@ -15,7 +15,7 @@
 
 BACKEND_2_DIFFERENTIATION = {
     "pytorch": "PSR",
-    "tensorflow": None,
+    "tensorflow": "PSR",
```
BrunoLiegiBastonLiegi, line 18:
```suggestion
    "tensorflow": None,
```

I would say that by default `keras` interface will use the native `tf` differentiation rather than `PSR`. In any case, the automatic selection of the differentiation method is likely to be changed in #42

src/qiboml/models/keras.py
```diff
@@ -33,29 +33,42 @@ def __post_init__(self):
 
         params = [p for param in self.circuit.get_parameters() for p in param]
         params = tf.Variable(self.backend.to_numpy(params))
-        self.circuit_parameters = self.add_weight(shape=params.shape, trainable=True)
-        self.set_weights([params])
+
+        self.circuit_parameters = self.add_weight(
+            shape=params.shape, initializer="random_normal", trainable=True
+        )
```
BrunoLiegiBastonLiegi, line 39: You don't want to randomly initialized the parameters actually, but rather use the exact same parameters of the layers you built the model with, for this reason there was a `set_weights` in the first place.

src/qiboml/models/keras.py
```diff
@@ -33,29 +33,42 @@ def __post_init__(self):
...
+            )
 
         else:
-            self.circuit.set_parameters(self.get_weights()[0])
-            # self.circuit.set_parameters(self.circuit_parameters)
-            x = self.encoding(x) + self.circuit
-            x = self.decoding(x)
 
-        return x
+            weights = tf.identity(self.circuit_parameters)
```
BrunoLiegiBastonLiegi, line 54: why do you need the `tf.identity`?

src/qiboml/models/keras.py
```diff
@@ -33,29 +33,42 @@ def __post_init__(self):
...
+            self.circuit.set_parameters(weights)
+
+            output = self.decoding(self.encoding(x) + self.circuit)
+            output = tf.expand_dims(output, axis=0)
```
BrunoLiegiBastonLiegi, line 58: are you sure you always need to expand the dims for each decoding?

src/qiboml/models/keras.py
```diff
@@ -33,29 +33,42 @@ def __post_init__(self):
...
         self,
     ):
         return self.output_shape
 
+    def draw(
+        self,
+    ):
+        breakpoint()
+        print("ciao")
+
```
BrunoLiegiBastonLiegi, line 71:
```suggestion
```

tests/Useless.py
```diff
@@ -0,0 +1,29 @@
...
+c.add(gates.RX(0, theta=0.4))
+
+random_choice = random_subset(nqubits, dim)
+print(f"Scelta random {random_choice}")
+result = c().probabilities()
+print(result)
+
+
+tensor = tf.random.uniform((2, nqubits), minval=0, maxval=2, dtype=tf.int32)
+print(f"Tensore: {tensor}")
```
BrunoLiegiBastonLiegi, line 29: if this test covers some case that is not tested in `tests/` then better to add it there, otherwise we should remove this

tests/conftest.py
```diff
@@ -10,12 +10,12 @@
 # backends to be tested
 BACKENDS = [
     "tensorflow",
-    "pytorch",
-    "jax",
+    # "pytorch",
+    # "jax",
```
BrunoLiegiBastonLiegi, line 14:
```suggestion
    "pytorch",
    "jax",
```
reactivate these to check that `pytorch` and `jax` are still working

tests/conftest.py
```diff
@@ -10,12 +10,12 @@
...
 ]
 
 FRONTENDS = [
-    "pytorch",
+    # "pytorch",
```
BrunoLiegiBastonLiegi, line 18:
```suggestion
    "pytorch",
```
same here

tests/test_models_interfaces.py
```diff
@@ -37,11 +38,31 @@ def random_subset(nqubits, k):
...
+        )
+    )
+    model.add(q_model)
+    model.add(
+        frontend.keras.layers.Dense(
+            12,
+            name="Dense1",
+            activation="relu",
+        )
+    )
```
BrunoLiegiBastonLiegi, line 58: what's this for? can't you use the `build_sequential` below?

BrunoLiegiBastonLiegi:
@MatteoRobbiati the ``keras`` interface is now working with the ``tensofrlow`` backend and differentiation. The ``PSR`` rule is still not working with ``keras``, but I would address that in a separate PR.

MatteoRobbiati (approved):
Thanks!
I tested it with a simple example and I was able to train and fit my function.
I found it working better with `GradientTape` than with the built in `mode.fit` method.. but ig this is something we can investigate.

PR #59, opened 2025-01-20
Adapt tutorial to recent qiboml updates

MatteoRobbiati:
As per title.

Modifications description:
- update set_backend call according to backend-platform paradigm;
- update interface import after migrating interfaces into their folder.


BrunoLiegiBastonLiegi (approved):
Thanks, this looks good. I just have two minor requests:
- add a semicolumn ( ; ) after each pip install in the first cell, this should hide the install process thet takes a lot of space
- remove the `nshots=None` from the decoding, as `None` should be the default value anyway and this may confuse the reader

BrunoLiegiBastonLiegi:
Not sure why but I get a 
``` python
ModuleNotFoundError: No module named 'qiboml.interfaces'
```
even though the notebook should be installing main, and we have `qiboml.interfaces` in main already.
Does it work for you?

MatteoRobbiati:
> Not sure why but I get a
> 
> ```python
> ModuleNotFoundError: No module named 'qiboml.interfaces'
> ```
> 
> even though the notebook should be installing main, and we have `qiboml.interfaces` in main already. Does it work for you?

I tested it just before pushing, and in my case it worked!

PR #60, opened 2025-01-21
Testing some improvements for the `JaxBackend`

BrunoLiegiBastonLiegi:
This tries to slightly speed up the `JaxBackend`. 

For comparison, considering this simple circuit:
``` python
0: ─RY─RZ─o─X─
1: ─RY─RZ─X─|─
2: ─RY─RZ───o─
```
with phase encoding and probabilities decoding, for 1000 repetitions I get:

   operation       |        main                      |       this PR                           |
-----------------------|--------------------------------|------------------------------------|
apply CNOT       | 0.4603139969985932 | 0.20405673499772092   |
execute_circuit |  6.132840211001167   | 4.305133709000074       |
forward              | 9.22812958600116     | 6.386330481000186        |
backward           | 0.7384055000002263 |  0.7501623499993002     |

which accounts for a ~30% reduction in the time to execute a circuit and compute a forward pass.


src/qiboml/backends/jax.py
```diff
@@ -88,6 +136,20 @@ def update_frequencies(self, frequencies, probabilities, nsamples):
         frequencies = frequencies.at[res].add(counts)
         return frequencies
 
+    def apply_gate(self, gate, state, nqubits):
+        if gate.is_controlled_by:
```
MatteoRobbiati, line 140: I think this condition will be satisfied only when the gate is built using the `controlled_by` method. For example, in the following

```python
from qibo import gates

cx = gates.CNOT(0,1)
print(cx.is_controlled_by)   # This is False
```

the gate is not satisfying the condition. If you want to use it on any controlled gates, I would use `len(gate.control_qubit) >= 1` instead (or we have to update Qibo).

MatteoRobbiati (commented):
Thanks! Just a comment follows.

src/qiboml/backends/jax.py
BrunoLiegiBastonLiegi: yes, the standard two qubit (or generically controlled) gates are executed by the ``__apply_gate`` below indeed

MatteoRobbiati (approved):
Thanks!

PR #62, opened 2025-01-27
Various fixes to the differentiation module

BrunoLiegiBastonLiegi:
This implements some fixes to differentiation:
- Enables gradient wrt input for PSR
- Provides encoders with a ``differentiable`` property
- Improves ``Jax`` by using ``functools.partial`` instead of class attributes to avoid leakage

### DIFF
--- a/src/qiboml/operations/differentiation.py
+++ b/src/qiboml/operations/differentiation.py
@@ -144,11 +144,7 @@ class Jax(Differentiation):
 
     def __init__(self):
         self._jax: Backend = JaxBackend()
-        # self._encoding = None
-        # self._training: Circuit = None
-        # self._decoding: QuantumDecoding = None
         self._argnums: tuple[int] = None
-        # self._circuit = None
 
     def evaluate(
         self,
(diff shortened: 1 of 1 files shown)


src/qiboml/models/encoding.py
```diff
@@ -32,8 +35,14 @@ def circuit(
     ):
         return self._circuit
 
+    @property
+    def differentiable(self):
```
MatteoRobbiati, line 39: Isn't maybe safer to set this property as False by default (1) or we assume cases like binary encoder are more exceptional while the default should in general be differentiable (2)?

src/qiboml/operations/differentiation.py
```diff
@@ -49,11 +59,24 @@ def evaluate(self, x: ndarray, encoding, training, decoding, backend, *parameter
...
+                            encoding,
+                            circuit,
+                            decoding,
+                        )
+                    ),
+                    (decoding.output_shape[-1], x.shape[-1]),
+                )
+            )
+        else:
+            gradient.append(backend.np.zeros((decoding.output_shape[-1], x.shape[-1])))
```
MatteoRobbiati, line 79: I am not sure we can just pad the gradient vector with values or zeros here. 
It is probably safe until the model structure is: encoding -> circuit -> decoding, but as soon as we will have a mix of trainable and non trainable gates we should probably loop over the gates. 
I say this because I guess the position of the gradients in the list is important, while here we are basically padding it at its beginning.

MatteoRobbiati (commented):
Thanks again @BrunoLiegiBastonLiegi for this!! 
I just added some comments below. 
Let's discuss them quickly and - once this is done - for me we are good to go.

src/qiboml/interfaces/pytorch.py
```diff
@@ -16,6 +16,7 @@
     "qiboml-pytorch": None,
     "qiboml-tensorflow": Jax,
     "qiboml-jax": Jax,
+    "numpy": Jax,
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: yes since ``qibo`` is a dependency, and ``numpy`` its native backend.

src/qiboml/models/encoding.py
BrunoLiegiBastonLiegi: I would say it's more common to have a differentiable encoding.

src/qiboml/interfaces/pytorch.py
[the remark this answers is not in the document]
MatteoRobbiati: > yes since `qibo` is a dependency, and `numpy` its native backend.

If Qibojit installed, the default should be Qibojit, no?

MatteoRobbiati (approved):

PR #66, opened 2025-02-25
`quantum_info` module refactor

BrunoLiegiBastonLiegi:
Fixes for the refactor qiboteam/qibo#1583

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13.5"]
     uses: qiboteam/workflows/.github/workflows/rules-poetry.yml@v1
     with:
       os: ${{ matrix.os }}
--- a/poetry.lock
+++ b/poetry.lock
@@ -1143,15 +1143,15 @@ woff = ["brotli (>=1.0.1) ; platform_python_implementation == \"CPython\"", "bro
 
 [[package]]
 name = "fsspec"
-version = "2025.9.0"
+version = "2025.10.0"
 description = "File-system specification"
 optional = false
 python-versions = ">=3.9"
 groups = ["main", "docs", "qibocal", "tests"]
 markers = "sys_platform == \"linux\" or sys_platform == \"darwin\" or sys_platform != \"linux\" and sys_platform != \"darwin\""
 files = [
-    {file = "fsspec-2025.9.0-py3-none-any.whl", hash = "sha256:530dc2a2af60a414a832059574df4a6e10cce927f6f4a78209390fe38955cfb7"},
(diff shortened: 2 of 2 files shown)


BrunoLiegiBastonLiegi:
@Edoardo-Pedicillo any idea why the docs workflow is complaining after I merged the main?

Edoardo-Pedicillo:
> @Edoardo-Pedicillo any idea why the docs workflow is complaining after I merged the main?

Yes, I forgot to update the triggers: the documentation workflow should be executed only when pushing on main to update the latest version and when a tag is generated to update the stable one. This is solved in [e25ca2a](https://github.com/qiboteam/qiboml/pull/66/commits/e25ca2a305d2901c28d9a1f1a6392150adcbbea9).

src/qiboml/backends/jax.py
```diff
@@ -24,6 +24,34 @@ def __init__(self):
...
+            pass
+
+        random = Random()
+        random.randint = lambda minval, maxval, size: self.jax.random.randint(
+            self.random_key, minval=minval, maxval=maxval, shape=size
+        )
+        random.uniform = lambda minval, maxval, size: self.jax.random.uniform(
+            self.random_key, minval=minval, maxval=maxval, shape=size
+        )
+        self.qinfo.ENGINE.random = random
```
renatomello, line 42: Is it really necessary to define a class inside another class?

src/qiboml/backends/jax.py
```diff
@@ -24,6 +24,34 @@ def __init__(self):
...
+
+        # load some custom qinfo operators
+        from qiboml.quantum_info.quantum_info_jax import QINFO
```
renatomello, line 45: Does it create a circular import error if this is placed at the top level?

src/qiboml/backends/pytorch.py
```diff
@@ -64,7 +65,29 @@ def __init__(self):
...
+        self.qinfo.ENGINE.random.normal = lambda loc, scale, size: torch.normal(
+            torch.tensor(loc).repeat(size), torch.tensor(scale).repeat(size)
+        )
+        self.qinfo.ENGINE.random.standard_normal = lambda size: torch.normal(
+            torch.zeros(size), torch.ones(size)
+        )
+        self.qinfo.ENGINE.random.uniform = (
+            lambda low, high, size: (low - high) * torch.rand(size) + high
+        )
+        self.qinfo.ENGINE.random.randint = torch.randint
```
renatomello, line 83: This seems to be a cleaner solution for the same problem that was also in the jax backend (with the class Random being defined inside the backend)

renatomello (commented):
It seems to me that the entire `quantum_info` module here in `qiboml` could be hidden from users since it is just necessary for compatibility.

src/qiboml/backends/jax.py
BrunoLiegiBastonLiegi: In this case I think I couldn't overwrite the function from jax directly as jax already has a random module. Thus, something like:
``` python
self.qinfo.ENGINE.random.randint = lambda minval, maxval, size: self.jax.random.randint(...
```
would result in an infinite loop. What I could do is backing up the original jax function first, but honestly I found cleaner to just group everything under a custom Random class.

BrunoLiegiBastonLiegi:
> It seems to me that the entire `quantum_info` module here in `qiboml` could be hidden from users since it is just necessary for compatibility.

what do you mean precisely?

Issue #63, opened 2025-02-07
are pytorch and tensorflow really using the GPU?


BrunoLiegiBastonLiegi:
Both the backends set the GPU as default device, if available:

https://github.com/qiboteam/qiboml/blob/96f7c471286192ef0e43295e811878c57d3acafe/src/qiboml/backends/pytorch.py#L55

https://github.com/qiboteam/qiboml/blob/96f7c471286192ef0e43295e811878c57d3acafe/src/qiboml/backends/tensorflow.py#L66C1-L70C46

however, if you try something as simple as:

``` python
from qibo import set_backend
from qibo.models import QFT

set_backend(backend="qiboml", platform="pytorch")
c = QFT(5)
r = c()
r.probabilities().device
# device(type='cpu')
r.state().device
# device(type='cpu')
```

you quickly notice that the GPU was probably not used (the same happens for tf).
At this point we might want to double check that the selected device is used in the computation.

PR #72, opened 2025-03-10
Enabling GPUs with the pytorch backend

BrunoLiegiBastonLiegi:
As per title, this tries to make generically device aware the pytorch backend. Depending on how complicate it is to do the same for jax and tensorflow, I may address those two here as well.

Adrresses #63 #76

EDIT: the GPU appears to be working fine with pytorch now (at least for what concerns ``qiboml``,  broader compatibility with the whole ``qibo`` ecosystem has still to be double checked). I would postpone to another PR the same patch to be applied to jax and tf

### DIFF
--- a/src/qiboml/backends/pytorch.py
+++ b/src/qiboml/backends/pytorch.py
@@ -71,6 +71,18 @@ class PyTorchBackend(NumpyBackend):
         self.np.sign = self.np.sgn
         self.np.flatnonzero = lambda x: self.np.nonzero(x).flatten()
 
+        # These functions are device dependent
+        torch_zeros = self.np.zeros
+
+        def zeros(shape, dtype=None, device=None):
+            if dtype is None:
+                dtype = self.dtype
+            if device is None:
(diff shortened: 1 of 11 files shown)


BrunoLiegiBastonLiegi:
@scarrazza does ``qiboml`` need to be allowed to run on GPUs? I remember we encountered something similar with ``qibojit``.

src/qiboml/backends/pytorch.py
```diff
@@ -15,20 +15,28 @@ class TorchMatrices(NumpyMatrices):
         dtype (torch.dtype): Data type of the matrices.
     """
 
-    def __init__(self, dtype):
+    def __init__(self, dtype, device):
```
renatomello, line 18: Maybe it is worth defaulting `device` to `None` and globally setting it to either GPU or CPU using https://pytorch.org/docs/stable/generated/torch.set_default_device.html

renatomello (approved):

PR #73, opened 2025-03-12
Some fixes to the pyproject and the publish workflow

BrunoLiegiBastonLiegi:
As per title.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,8 +14,8 @@ documentation = "https://qibo.science/docs/qiboml/stable"
 packages = [{ include = "qiboml", from = "src" }]
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.13"
-numpy = "^1.26.4"
+python = ">=3.10,<3.13"
+numpy = "^2.0.0"
 keras = { version = "^3.0.0", optional = true }
 tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux' or sys_platform == 'darwin'", optional = true }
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [ 3.9, '3.10', '3.11', '3.12']
+        python-version: ['3.10', '3.11', '3.12']
```
renatomello, line 17: Will we introduce `3.13` after the `qibo` release? Currently `pip` complains about it when installing `qibo` from source.

renatomello: oh yeah you're right. I forgot about that. Sorry.

pyproject.toml
```diff
@@ -59,28 +59,41 @@ keras = { version = "^3.0.0", optional = true }
 tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux' or sys_platform == 'darwin'"}
 torch = { version = "^2.3.1"}
 
-[tool.poetry.group.benchmark.dependencies]
-pytest-benchmark = { version = "^4.0.0", extras = ["histogram"] }
+#[tool.poetry.group.benchmark.dependencies]
+#pytest-benchmark = { version = "^4.0.0", extras = ["histogram"] }
+
```
Edoardo-Pedicillo, line 64: Remove commented lines.

PR #79, opened 2025-03-21
Add ML frameworks to docs deps

MatteoRobbiati:
As per title.


BrunoLiegiBastonLiegi:
@MatteoRobbiati I am not sure what ``pylint`` is complaining about, anyway can I also ask you to point back to ``qibo`` main here:
https://github.com/qiboteam/qiboml/blob/e907731c2a7b6d8a8aadd11e903ca59bf308be7a/pyproject.toml#L24

because the corresponding PR has just been merged.

alecandido:
@BrunoLiegiBastonLiegi given today's release, you may actually use that one, not even `main` (until you'll need that again - hopefully never, but if it ought to be...)

MatteoRobbiati:
Everything should be fixed now.
Sorry for that, but I also did a bit of cleaning by removing some unused files/folders.

PR #89, opened 2025-05-13
Enable ``density_matrix=True`` execution for noisy training

BrunoLiegiBastonLiegi:
As per title, this adds a ``density_matrix`` argument to both encoders and ansaetze to enable density matrix simulations. In particular this is tested for noisy training. In order to make that work I had to implement custom ``matrix_fused`` methods for the backends, as in the numpy backend the gate was converted to numpy to use the scipy sparse csr representation. Unfortunately, jax and tensorflow only support the coo representation, whereas pytorch supports also csr if ``mkl`` is installed, thus at the moment I switched to the coo representation. However, I am not sure whether it makes sense and if wouldn't be better to just do everything with dense array at this point, which in our case may also end up being faster than coo ones.

EDIT: a quick test with two qubits showed as jax is way slower using coo arrays rather than dense, no major differences for torch and tensorflow instead. Therefore, for now, I am keeping sparse representation with tensorflow and torch (if mkl is available) and always using dense arrays with jax.


BrunoLiegiBastonLiegi:
@stavros11 @alecandido may I ask your opinion on the whole csr vs coo vs dense matter?

alecandido:
> @stavros11 @alecandido may I ask your opinion on the whole csr vs coo vs dense matter?

Well, you know that the usage of sparse matrices truly depend on the application, since it's driven by the ratio of non-zero values. The optimal layout possibly depends even on the distribution of these non-zero values within the matrix, and certainly on the operations you may want to perform on it.

Beyond the useless premise, you can check yourself on [Wikipedia](https://en.wikipedia.org/wiki/Sparse_matrix) which are the layouts, and advantages and disadvantages. Or even SciPy docs, which have a quick [summary about the formats](https://scipy.github.io/devdocs/tutorial/sparse.html#understanding-sparse-array-formats).

In the context of Qibo, I'm not even sure of all places where they are used, but I expect that for parametrized gates and especially QML the number of non-zero elements in the computational basis may drastically drop. So, I largely expect that in this case they are of little to no use (over a certain threshold of non-zero, sparse matrices are clearly a huge waste...).

In any case, I would even hope they are somehow optional. In which case there would be nothing bad in keeping them as optional, and just leave the error natively raised by unsupported operations.

All in all, I do not have a strong opinion, since I should first investigate their usage.

src/qiboml/backends/tensorflow.py
```diff
@@ -128,9 +128,70 @@ def matrix_parametrized(self, gate):
         npmatrix = super().matrix_parametrized(gate)
         return self.tf.cast(npmatrix, dtype=self.dtype)
 
-    def matrix_fused(self, gate):
-        npmatrix = super().matrix_fused(gate)
-        return self.tf.cast(npmatrix, dtype=self.dtype)
+    def block_diag(self, *mats):
```
MatteoRobbiati, line 131: I don't know if you decided not to use [tf.linalg.LinearOperatorBlockDiag](https://www.tensorflow.org/api_docs/python/tf/linalg/LinearOperatorBlockDiag) on purpose or not, but I'll paste it here in case you want to have a look.

src/qiboml/backends/tensorflow.py
```diff
@@ -128,9 +128,70 @@ def matrix_parametrized(self, gate):
...
+            qubits = list(gate.qubits)
+            indices = qubits + [q for q in fgate.target_qubits if q not in qubits]
+            indices = np.argsort(indices)
+            transpose_indices = list(indices)
+            transpose_indices.extend(indices + rank)
+            gmatrix = self.np.transpose(gmatrix, transpose_indices)
+            gmatrix = self.np.reshape(gmatrix, original_shape)
+            # fuse the individual gate matrix to the total ``FusedGate`` matrix
+            # we are using sparse matrices to improve perfomances
+            matrix = self.tf.sparse.sparse_dense_matmul(
```
MatteoRobbiati, line 190: We should probably check if the performances are improved in the case of an high number of qubits. Anyway, LGTM.

src/qiboml/interfaces/pytorch.py
```diff
@@ -179,7 +179,9 @@ def forward(
 
         # Build the temporary circuit from the circuit structure.
         ctx.differentiable_encodings = True
-        circuit = Circuit(decoding.nqubits)
+        circuit = Circuit(
+            decoding.nqubits, density_matrix=circuit_structure[0].density_matrix
```
MatteoRobbiati, line 183: In the `_uniform_circuit_structure_density_matrix` function we could probably also return the boolean value and set it as an attribute of the model class. On this way, we can use that attribute here directly (improving a bit the readability).

src/qiboml/interfaces/utils.py
```diff
@@ -99,3 +100,9 @@ def draw_circuit(circuit_structure, backend, plt_drawing=True, **plt_kwargs):
     else:
         circuit.draw()
         return str(circuit)
+
+
+def _uniform_circuit_structure_density_matrix(circuit_structure):
```
MatteoRobbiati, line 105:
```suggestion
def _uniform_circuit_structure_density_matrix(circuit_structure):
    """
    Align the ``density_matrix`` attribute of all circuits composing the circuit structure.
    Namely, setting them to ``True`` if at least one component of the circuit has ``density_matrix==True``.
    """
```

MatteoRobbiati (commented):
Thanks! A few comments follow.

src/qiboml/backends/tensorflow.py
BrunoLiegiBastonLiegi: Thanks I couldn't find it

src/qiboml/interfaces/pytorch.py
BrunoLiegiBastonLiegi: Not sure it is worth making it an attribute of the ``QuantumModel``, maybe a property similarly to what we do with ``backend``, but even so you would need to modify ``circuit_from_structure`` to take as input ``density_matrix``. All in all, I think it's just simpler to extract it from the first component of circuit structure, this is a method the user should not use anyway.

src/qiboml/backends/tensorflow.py
BrunoLiegiBastonLiegi: with 6 qubits there is virtually no difference as well

PR #94, opened 2025-05-19
Transpiler support for qibolab execution

BrunoLiegiBastonLiegi:
This adds the option to pass a transpiler in the decoder to allow for execution through qibolab especially.
A test using qibolab's emulator is added.


tests/emulator/calibration.json
andrea-pasquale, line 1: This file is needed only if you create the platform from qibocal using `create_calibration_platform`. Since you are using `create_platform` from qibolab it should not be necessary.

src/qiboml/models/encoding.py
```diff
@@ -39,10 +39,12 @@ def __post_init__(
 
     @cached_property
     def _data_to_gate(self):
-        raise_error(
-            NotImplementedError,
-            f"_data_to_gate method is not implemented for encoding {self}.",
-        )
+        """
```
MatteoRobbiati, line 42: Are we sure we want this as default for any encoding? I think it is safer to raise a `NotImplementedError` to avoid custom encoders to inherit wrong features.

pyproject.toml
```diff
@@ -62,6 +62,7 @@ matplotlib = "^3.7.0"
 keras = "^3.0.0"
 tensorflow = { version = "^2.16.1", markers = "sys_platform == 'linux' or sys_platform == 'darwin'"}
 torch = { version = "^2.3.1"}
+#qibolab = { git="https://github.com/qiboteam/qibolab", extras=["emulator"] }
```
MatteoRobbiati, line 65: We need to uncomment this, right?

MatteoRobbiati (changes requested):
Thanks Andre, I started to have a look here (maybe too early, sorry for that 😄)

src/qiboml/models/encoding.py
BrunoLiegiBastonLiegi: I think it would make sense to have a default value for this, as long as we make it clear what is the default value and specify that you need to define your custom one if the default does not fit your needs. I mostly did this to avoid repeating the same ``_data_to_gate`` in the ``BinaryEncoding`` and possibly in all the other simple encoding that simply map the ``i``-th input to the ``i``-th gate. But anyway, it's not really necessary, we can avoid this if you don't like it.

MatteoRobbiati: Well, ok. Indeed, if it is documented, then the error is responsibility of the users. I am fine with it, I was trying to anticipate users' errors, which is indeed a non-maintainable approach :)

pyproject.toml
alecandido: We already discussed a solution with @scarrazza and @stavros11. It will be delivered soon.

However, notice that the solution will consist in limiting the dependencies, getting rid of optional ones as much as possible. This is something that should be considered also in `qiboml` (which starts proliferating, and it will lead to very similar problems).

MatteoRobbiati (approved):
Thanks! Just a couple of minor comments follow.

src/qiboml/models/encoding.py
```diff
@@ -39,10 +39,17 @@ def __post_init__(
...
+        Mapping between the index of the input and the indices of the gates in the
+        produced encoding circuit queue, where the input is encoded to.
+        For instance, {0: [0,2], 1: [2,]}, represents an encoding where the element
```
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Not really, just to stress that it is a list, I personally like it more, but it is up to debate.

PR #95, opened 2025-05-22
Improve README

MatteoRobbiati:
Improve README structure adding details and a minimum working example (VQE).


src/qiboml/models/ansatze.py
```diff
@@ -13,7 +13,7 @@ def HardwareEfficient(
     circuit = Circuit(nqubits)
 
     for _ in range(nlayers):
-        for q in qubits:
+        for q in range(qubits):
```
BrunoLiegiBastonLiegi, line 16:
```suggestion
        for q in qubits:
```
this is actually just qubits, because qubits is a tuple already

PR #96, opened 2025-05-23
Allow for no encoder and no inputs to be passed to the ``QauntumModel``

BrunoLiegiBastonLiegi:
As per title, useful for implementing VQEs.


renatomello:
@BrunoLiegiBastonLiegi you're a genius

scarrazza:
@BrunoLiegiBastonLiegi are you sure the tests for linux are passing locally? Given that main is passing, the crashes-slow down seems related to changes made in this PR.

PR #104, opened 2025-05-29
Enabling ``wire_names``

BrunoLiegiBastonLiegi:
This gives the possibility to set the ``wire_names`` in encoders and ansatze through their ``qubits`` argument. This is mostly useful for selecting the qubits to use when running on hardware.

### DIFF
--- a/src/qiboml/models/decoding.py
+++ b/src/qiboml/models/decoding.py
@@ -54,6 +54,9 @@ class QuantumDecoding:
         """
         self._circuit.density_matrix = x.density_matrix
         self._circuit.init_kwargs["density_matrix"] = x.density_matrix
+        self._circuit.wire_names = x.wire_names
+        self._circuit.init_kwargs["wire_names"] = x.wire_names
+
         if self.transpiler is not None:
             x, _ = self.transpiler(x)
         return self.backend.execute_circuit(x + self._circuit, nshots=self.nshots)
--- a/src/qiboml/models/encoding.py
+++ b/src/qiboml/models/encoding.py
@@ -2,13 +2,14 @@ import inspect
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from functools import cached_property
-from typing import Optional
+from typing import Optional, Union
 
 import numpy as np
 from qibo import Circuit, gates
 from qibo.config import raise_error
 
 from qiboml import ndarray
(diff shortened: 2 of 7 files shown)


sorewachigauyo:
Waiting for https://github.com/qiboteam/qibolab/issues/1226 to be sorted out

src/qiboml/models/encoding.py
```diff
@@ -18,23 +19,25 @@ class QuantumEncoding(ABC):
 
     Args:
         nqubits (int): total number of qubits.
-        qubits (tuple[int], optional): set of qubits it acts on, by default ``range(nqubits)``.
+        qubits (tuple[int] | tuple[str], optional): set of qubits it acts on, by default ``range(nqubits)``. Optionally, the name of the wires to run on can be passed through this argument. This is mainly useful when executing on hardware to select which qubits to make use of.
```
renatomello, line 22:
```suggestion
        qubits (tuple[int] | tuple[str], optional): set of qubits it acts on, by default ``range(nqubits)``. 
            Optionally, the name of the wires to run on can be passed through this argument. 
            This is mainly useful when executing on hardware to select which qubits to make use of.
```

src/qiboml/models/decoding.py
```diff
@@ -19,6 +20,8 @@ class QuantumDecoding:
     Args:
         nqubits (int): total number of qubits.
         qubits (tuple[int], optional): set of qubits it acts on, by default ``range(nqubits)``.
+            Optionally, the name of the wires to run on can be passed through this argument.
```
MatteoRobbiati, line 23: Maybe I would add a small example to explain how to pass the argument in case we also want to set the `wire_names`.

src/qiboml/models/encoding.py
```diff
@@ -18,7 +19,7 @@ class QuantumEncoding(ABC):
 
     Args:
         nqubits (int): total number of qubits.
-        qubits (tuple[int], optional): set of qubits it acts on, by default ``range(nqubits)``.
+        qubits (tuple[int] | tuple[str], optional): set of qubits it acts on, by default ``range(nqubits)``.
```
MatteoRobbiati, line 22: Is this necessary now that we moved the `wire_names` passing to the decoders?

MatteoRobbiati (commented):
Thanks! Just a couple of extra comments follow.

PR #105, opened 2025-05-29
Real time quantum error mitigation

MatteoRobbiati:
Here, a real-time error mitigation algorithm is implemented. 
The method relies on the `Mitigator` class and it is only available for the `Expectation` decoder.


src/qiboml/models/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    def __post_init__(self):
+        self.backend = _check_backend(self.backend)
+        self._mitigation_map: Callable[..., ndarray] = lambda x, a=1, b=0: a * x + b
+
+        cfg = self.mitigation_config or {}
+        self._real_time_mitigation = cfg.get("real_time", False)
+        self._mitigation_method = cfg.get("method", "cdr")
+        self._mitigation_method_kwargs = cfg.get("method_kwargs", {})
+
+        custom_map = cfg.get("method_kwargs", {}).get("model")
```
BrunoLiegiBastonLiegi, line 38:
```suggestion
        custom_map = self._mitigation_method_kwargs.get("model")
```

src/qiboml/models/utils.py
```diff
@@ -0,0 +1,70 @@
...
+            circuit=circuit,
+            observable=observable,
+            noise_model=noise_model,
+            nshots=nshots,
+            full_output=True,
+            **self._mitigation_method_kwargs,
+        )
+
+        self._mitigation_map.__defaults__ = tuple(popt)
+        self._mitigation_map_popt = self.backend.np.array(popt)
```
BrunoLiegiBastonLiegi, line 69:
```suggestion
        self._mitigation_map_popt = self.backend.cast(popt, dtype="double")
```

tests/test_mitigation.py
```diff
@@ -0,0 +1,163 @@
...
+        backend=backend,
+        noise_model=build_noise_model(nqubits=nqubits, local_pauli_noise_prob=0.04),
+        density_matrix=True,
+        mitigation_config=mitigation_config,
+    )
+
+    mit_model = frontend.QuantumModel(
+        circuit_structure=deepcopy(vqe),
+        decoding=mit_decoding,
+        differentiation=PSR(),
```
BrunoLiegiBastonLiegi, line 109: we should test with all the other differentiation methods as well

tests/test_mitigation.py
```diff
@@ -0,0 +1,163 @@
...
+        frontend=frontend,
+        backend=backend,
+        model=mit_model,
+        epochs=30,
+    )
+
+    assert mit_result < noisy_result
+
+
+def test_custom_map(frontend):
```
BrunoLiegiBastonLiegi, line 122: this test could probably be removed by adding a ``pytest`` parameter ``noise_map`` to the previous test

src/qiboml/models/decoding.py
```diff
@@ -53,10 +57,21 @@ def __call__(
...
+            self._circuit.init_kwargs["density_matrix"] = x.density_matrix
+
         if self.transpiler is not None:
             x, _ = self.transpiler(x)
-        return self.backend.execute_circuit(x + self._circuit, nshots=self.nshots)
+
+        if self.noise_model is not None:
+            executable_circuit = self.noise_model.apply(x + self._circuit)
+        else:
+            executable_circuit = x + self._circuit
```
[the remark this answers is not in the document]
MatteoRobbiati: I think it makes more sense to apply the noise to the transpiled one, since it is actually what is happening on the device.

tests/test_mitigation.py
MatteoRobbiati: I decided to test PSR only because imo it makes sense to take into account mitigation whenever we run on hardware or with shots, and so with PSR. Would it make sense to use noisy simulation in exact mode? (it can, in fact. Mine is more a discussion here)

tests/test_mitigation.py
MatteoRobbiati: I would keep it separate just because this one is very short (2 epochs), and I would anyway break the pipeline of the one above. But well, I am fine with changing it if you prefer :)

src/qiboml/models/decoding.py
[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: I mean, I don't have any particular strong opinion, as long as we make it clear in the docstring.

tests/test_mitigation.py
BrunoLiegiBastonLiegi: No particularly strong opinion here as well, I was mainly pointing this out because I noticed things that I was would have failed in other settings, e.g. pytorch native.

src/qiboml/models/decoding.py
```diff
@@ -157,36 +181,83 @@ class Expectation(QuantumDecoding):
...
             (ndarray): the calculated expectation value.
         """
+        # recompute map if real-time enabled and not yet run
+        if (
+            self.mitigation_config is not None
+            and self.mitigator._real_time_mitigation
+            and self.backend.np.allclose(
+                self.mitigator._mitigation_map_popt,
+                self.mitigator._mitigation_map_initial_popt,
+                atol=1e-6,
```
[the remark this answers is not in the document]
MatteoRobbiati: No, this is just an initial update of the map and to check whether I never computed the RTQEM I am checking if the params are == to the default. 
In practice: there are default values, if still default, update. But the rtqem strategy (I mean how to trigger the rtqem) can be probably re-thought.

tests/test_interfaces.py
```diff
@@ -307,214 +296,215 @@ def backprop_test(frontend, model, data, target):
...
+
+# @pytest.mark.parametrize("layer,seed", zip(DECODING_LAYERS, [1, 3, 1, 26]))
+# def test_decoding(backend, frontend, layer, seed):
+#     if not layer.analytic and not layer is dec.Expectation:
+#         pytest.skip(
+#             "Expectation layer is the only differentiable decoding when the diffrule is not analytical."
+#         )
+
+#     set_device(frontend)
+#     set_seed(frontend, seed)
```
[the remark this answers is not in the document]
MatteoRobbiati: Sorry, I was debugging. Restoring them soon.

src/qiboml/models/decoding.py
```diff
@@ -157,36 +182,137 @@ class Expectation(QuantumDecoding):
...
+        # If mitigation is requested
+        if self.mitigation_config is not None:
+            # Construct the Mitigator object
+            self.mitigator = Mitigator(
+                mitigation_config=self.mitigation_config,
+                backend=self.backend,
+            )
+            # This attribute will contain the reference circuit (Clifford and
+            # error sensitive, that will be used in the real time error mitigation).
+            self._mitigation_reference_circuit = None
```
BrunoLiegiBastonLiegi, line 230: can we move this to the ``Mitigator`` object directly?

src/qiboml/models/decoding.py
```diff
@@ -157,36 +182,137 @@ class Expectation(QuantumDecoding):
...
+        Compute the mitigated expectation value using the saved map and a given
+        noisy expectation value.
+        """
+        mit_expval = self.backend.cast(
+            self.mitigator._mitigation_map(
+                expval, *self.mitigator._mitigation_map_popt
+            ),
+            dtype=self.backend.np.float64,
+        )
+        return mit_expval
```
BrunoLiegiBastonLiegi, line 275: this also seems more a method of ``Mitigator`` rather than Expectation, something like ``Mitigator.apply`` or ``Mitigator.__call__``, I am saying this because if in the future we will need mitigation in another decoder we won't need to rewrite these methods.

src/qiboml/models/decoding.py
```diff
@@ -157,36 +182,137 @@ class Expectation(QuantumDecoding):
...
+        if (
+            abs(mit_expval - self._mitigation_reference_value)
+            > self.mitigator._threshold
+        ):
+            self.mitigator.data_regression(
+                circuit=x + self._circuit,
+                observable=self.observable,
+                noise_model=self.noise_model,
+                nshots=self.nshots,
+            )
```
BrunoLiegiBastonLiegi, line 315: This as well would be more suited to be inside of ``Mitigator`` directly, in my opinion

src/qiboml/models/decoding.py
```diff
@@ -157,36 +182,137 @@ class Expectation(QuantumDecoding):
...
+        The check is performed according to https://arxiv.org/abs/2311.05680.
+        """
+        if self._mitigation_reference_circuit is None:
+            # Sample the reference error sensitive circuit (it will be pure Clifford)
+            self._mitigation_reference_circuit = error_sensitive_circuit(
+                circuit=x,
+                observable=self.observable,
+            )[0]
+            # TODO: replace Numpy with Clifford after fixing Unitary problem
+            simulation_backend = NumpyBackend()
```
BrunoLiegiBastonLiegi, line 290: this for instance would be an attribute of the mitigator ``self.simulation_backend``

src/qiboml/models/utils.py
```diff
@@ -1,4 +1,82 @@
...
+
+        custom_map = self._mitigation_method_kwargs.get("model")
+        if custom_map is not None:
+            if not callable(custom_map):
+                raise ValueError("Noise map model must be a callable")
+            self._mitigation_map = custom_map
+
+        n_params = self._mitigation_map.__code__.co_argcount - 1
+        defaults = self._mitigation_map.__defaults__ or tuple(
+            [1.0] * (n_params - 1) + [0.0]
```
BrunoLiegiBastonLiegi, line 48: is this better than random uniform or random normal initialization?

src/qiboml/models/utils.py
```diff
@@ -1,4 +1,82 @@
...
+        circuit: Circuit,
+        observable: Union[ndarray, Hamiltonian],
+        noise_model: NoiseModel,
+        nshots: Optional[int],
+    ):
+        """
+        Perform data regression on noisy and exact data.
+        """
+
+        _, _, popt, _ = getattr(error_mitigation, self._mitigation_method)(
```
BrunoLiegiBastonLiegi, line 64: this could be done in the post init directly, to avoid re-getting the same method over and over

BrunoLiegiBastonLiegi (commented):
Thanks, can you also remove this function that is no longer needed I presume:
https://github.com/qiboteam/qiboml/blob/6b56556432638c868dad3aa4a4ed8b4ad42d9500/src/qiboml/interfaces/utils.py#L106C1-L113C45

src/qiboml/models/utils.py
MatteoRobbiati: I personally prefer it, since this is the map $f(x)=x$.

PR #110, opened 2025-06-06
Bloch sphere

niccololaurora:
The pr contains an implementation of a Bloch sphere and a tutorial.

The Bloch sphere accepts both states (2D array) and vectors (3D array).
States and vectors can be plotted as arrows or as points: the "point representation" could be useful when numerous elements are present on the sphere (predictions of a training), whereas the "arrow representation" could be useful for representing few states (labels of classes).

If the Bloch sphere is fed an unnormalized state or a vector that does not lie on the sphere it gives an error.


renatomello:
Correct me if I'm wrong, but this seems more applicable to be in `qibo` than `qiboml`, no? Either in the `ui` module or the `quantum_info` module.

alecandido:
> Correct me if I'm wrong, but this seems more applicable to be in `qibo` than `qiboml`, no? Either in the `ui` module or the `quantum_info` module.

I would candidate it for the `ui` module

MatteoRobbiati:
I guess we can close this, since we opened an equivalent PR in Qibo, right? @niccololaurora @BrunoLiegiBastonLiegi

niccololaurora:
Yes! :)

Issue #103, opened 2025-05-28
Parameters initialisation


MatteoRobbiati:
Implement automatic parameters' initialisation: small angles, random, zeros, etc.

PR #111, opened 2025-06-10
Initialisation methods for the parameters of QuantumModel

niccololaurora:
This PR offers an easy initialisation for the parameters of the `QuantumModel` class.
The possible initialisation scenarios are (both for the Keras and Torch interface):

1. No initialisation triggers the default settings (gaussian init of the parameters); 
2. Initialisation with a np.ndarray (this method requires to be careful with the shape of the array);
3. Initialisation with tf.keras.initializers.Initializer for Keras and torch.nn.init for PyTorch (this method does not require to be careful with shapes);

There's a small tutorial on how to handle the initialisation, but I can show an example here (Keras):

`
q_model = QuantumModel(
    circuit_structure=[encoding, circuit],
    angles_initialisation= keras.initializers.RandomUniform(minval=-0.05, maxval=0.05, seed=None) ,
    decoding=decoding
)
`

This should address #103


MatteoRobbiati (changes requested):
Thanks for this! 
A few general comments:

1. I agree with using default initialisers from Keras and Pytorch;
2. Please add docstrings to the interfaces, so that the `angles_initialisation` is documented and users can understand how to play with the initialisers;
3. I would remove the tutorial, it is not really necessary to show a notebook for this. Instead, we can add a couple of lines into an already existing tutorial.
4. Tests are missing! Please update one of the tests (I suggest using the `test_models_interfaces`) passing the angles initialisation, or add a specific test (in the same file) to cover all the cases.

BrunoLiegiBastonLiegi (commented):
Thanks, this looks very nice, just a handful of comments. By the way, I would suggest to resolve the conflicts and merge the main here, because there have been major changes recently (they should not affect this PR much, but still better to doublecheck).

src/qiboml/interfaces/keras.py
```diff
@@ -216,6 +207,15 @@ def draw(self, plt_drawing=True, **plt_kwargs):
...
         return fig
 
+    def get_parameters(self) -> tf.Tensor:
+        """
+        Get the trainable parameters of the model.
+
+        Returns:
+            (tf.Tensor): the trainable parameters.
+        """
+        return self.circuit_parameters
```
BrunoLiegiBastonLiegi, line 217: is this really needed? `circuit_parameters` is already an attribute of the model so you can just call `QuantumModel.circuit_parameters`, but more often one will rely directly on the methods provided by keras/torch, i.e. `model.get_weights` and `model.parameters()`

src/qiboml/interfaces/pytorch.py
```diff
@@ -207,6 +201,15 @@ def draw(self, plt_drawing=True, **plt_kwargs):
...
 
+    def get_parameters(self) -> torch.Tensor:
+        """
+        Get the trainable parameters of the model.
+
+        Returns:
+            (torch.Tensor): the trainable parameters.
+        """
+        return self.circuit_parameters
+
```
BrunoLiegiBastonLiegi, line 212: same

tests/test_interfaces.py
```diff
@@ -625,6 +580,93 @@ def test_qibolab(frontend):
...
+            decoding=dec.Expectation(
+            nqubits=nqubits,
+            backend=backend,
+            )
+        )
+    
+    # Function to check the parameters
+    def assert_check(model_params, initializer):
+        if frontend.__name__ == "qiboml.interfaces.keras":
+            # assert np.all(np.equal(model_params.numpy(), initializer))
```
BrunoLiegiBastonLiegi, line 608:
```suggestion
```

tests/test_interfaces.py
```diff
@@ -625,6 +580,93 @@ def test_qibolab(frontend):
...
+            assert np.all(np.equal(model_params.detach().numpy(), initializer))
+
+    # Initializer: "tf.keras.initializers" or "torch.nn.init"
+    if with_initializer == True:
+        if frontend.__name__ == "qiboml.interfaces.keras":
+            frontend.tf.random.set_seed(1)
+            initializer = frontend.tf.keras.initializers.RandomNormal(mean=0.0, stddev=0.05)
+    
+            model = q_model(nqubits, initializer, circuit_structure)
+            model_params = model.get_parameters()
```
BrunoLiegiBastonLiegi, line 620: ah ok I see now why you defined the `get_parameters` above, but there is no need to have it in the model, there's an utility function for that defined above:
https://github.com/qiboteam/qiboml/blob/7017c1400156ff88322aa31aed3db23450096726/tests/test_interfaces.py#L256

Sorry about that, it's not very visible, we should probably separate all the test utilities in a `tests/utils.py` file.

tests/test_interfaces.py
```diff
@@ -625,6 +580,93 @@ def test_qibolab(frontend):
...
+            values = initializer(shape=model_params.shape)
+            assert_check(model_params, values)
+
+        elif frontend.__name__ == "qiboml.interfaces.pytorch":
+            frontend.torch.manual_seed(1)
+            initializer = lambda p: frontend.torch.nn.init.normal_(p, mean=0, std=0.05)
+
+            model = q_model(nqubits, initializer, circuit_structure)
+            model_params = model.get_parameters()
```
BrunoLiegiBastonLiegi, line 629: here as well.
```suggestion
            model_params = get_parameters(frontend, model, True)
```

tests/test_interfaces.py
```diff
@@ -625,6 +580,93 @@ def test_qibolab(frontend):
...
+
+            assert_check(model_params, values)
+        else:
+            raise_error(RuntimeError, f"Unknown frontend {frontend}.")
+
+        
+    # No initializer
+    elif with_initializer == False:
+        initializer = None
+        model = q_model(nqubits, initializer, circuit_structure)
```
BrunoLiegiBastonLiegi, line 643:
```suggestion
```
this probably is not needed as it is the standard case tested by all the other tests

BrunoLiegiBastonLiegi (commented):
Thanks, this is ready to go, just some very tiny adjustments.

BrunoLiegiBastonLiegi (approved):

Issue #108, opened 2025-06-04
QAOA with Qiboml


mho291:
Related to Issue #99:

Trying to run QAOA MaxCut by updating only `params = [gamma, beta]` but it seems that there are some issues with how the params are read.

This is the code I used:

``` python
# Build graph
G = nx.Graph()
G.add_edge(0, 1)
G.add_edge(0, 2)
G.add_edge(0, 3)
G.add_edge(1, 2)
G.add_edge(2, 3)

# Create class for maxcut
class build_maxcut_circ(Circuit):
    def __init__(self, nqubits=None, accelerators=None, density_matrix=False, wire_names=None, G=None, params=None):
        super().__init__(nqubits, accelerators, density_matrix, wire_names)
        self.G = G
        self.params = params
    
    def set_parameters(self, params):
        self.params = params
        self.nlayers = int(len(params)/2)

    def get_parameters(self):
        return self.params

    def build(self):
        self.nlayers = int(len(params)/2)
        self.beta = self.params[self.nlayers:]
        self.gamma = self.params[:self.nlayers]

        for _i in range(len(G.nodes())):
            self.add(gates.H(_i))
        
        for p in range(self.nlayers):
            for i,j in list(self.G.edges()):
                self.add(gates.CNOT(i, j))
                self.add(gates.RZ(j, theta=self.gamma[p], trainable=True))
                self.add(gates.CNOT(i, j))
    
            for _j in range(len(G.nodes())):
                self.add(gates.RX(_j, theta=self.beta[p], trainable=True))
        return self

# Input parameters
params = np.array([[0.1], [0.2]]) # torch.Tensor([[0.1], [0.2]])
C = build_maxcut_circ(nqubits=len(G.nodes()), G=G, params=params)
C.set_parameters(params)
circ = C.build()

# Constructing decoding 
from qibo.hamiltonians import SymbolicHamiltonian
from qibo.symbols import Z

# Construct the decoding layer
hamiltonian = 0
for i, j in list(G.edges()):
    hamiltonian += 0.5*1 - 0.5*np.prod([Z(i), Z(j)])

# Define Hamiltonian
hamiltonian = SymbolicHamiltonian(hamiltonian)
decoding = Expectation(
                nqubits=nqubits,
                observable=hamiltonian,
            )

# Use pytorch quantum model
model = QuantumModel(
    circuit_structure = [C,],
    decoding=decoding,
)

# Optimize
[...]

Issue #99, opened 2025-05-27
Support for shared or linked parameters in ansatze


jf-kong:
The current way of passing in `circuit_structure` and `decoding` to the `QuantumModel` does not directly support trainable ansatze with shared or linked parameters Such examples include QCNN, UCC ansatze etc, where trainable parameters in different gates are linked by some relationship.

Had a discussion with @scarrazza earlier today, and one way that we have attempted to make the QCNN compatible is to create a class that inherits `Circuit`, and override the `set_parameter` and `get_parameter` methods. This object can then be passed to `circuit_structure`. We have an example that shows that this works with the QCNN. Potential issues that can crop up down the road with this method could be: different `Circuit` objects being concatenated together, and `set/get_pararameters` do not function properly.

@MatteoRobbiati, may I check what are your thoughts on this?

MatteoRobbiati:
> The current way of passing in `circuit_structure` and `decoding` to the `QuantumModel` does not directly support trainable ansatze with shared or linked parameters Such examples include QCNN, UCC ansatze etc, where trainable parameters in different gates are linked by some relationship.

This is half correct. This feature is already supported in the case the training is performed in exact simulation mode and keeping aligned interface and backend. In practice, if the quantum model is defined such that the same trainable parameter is affecting more gates, Pytorch or Tensorflow should be already able to keep track of this pattern and train with this in "mind". To be completely honest, I never tried it personally, but this is what I would expect. 
For sure, if we then talk about parameter shift rules or custom differentiation in general, then it is 100% true: we don't support this right now. 

But, this is definitely a very important feature we need and desire to support! In fact, thanks a lot for opening the discussion @jf-kong. 

> Had a discussion with [@scarrazza](https://github.com/scarrazza) earlier today, and one way that we have attempted to make the QCNN compatible is to create a class that inherits `Circuit`, and override the `set_parameter` and `get_parameter` methods. This object can then be passed to `circuit_structure`. We have an example that shows that this works with the QCNN. Potential issues that can crop up down the road with this method could be: different `Circuit` objects being concatenated together, and `set/get_pararameters` do not function properly.

I am for sure curious of having a look at this! I would also suggest to try to implement the solution as mentioned before (which is maybe what you are already doing: setting manually the parameters inside the circuit with some custom rule).

@BrunoLiegiBastonLiegi, feel free to add any comment/correction here :)

PR #112, opened 2025-06-10
Equivariant models

BrunoLiegiBastonLiegi:
~~This allows for sharing parameters inter and intra circuit layers. Namely, the independent parameters are extracted from the circuit using the [np.shares_memory](https://numpy.org/doc/stable/reference/generated/numpy.shares_memory.html) method and added as parameters of the model. Upon circuit reconstruction, then, a new custom ``set_parameters`` function is used to duplicate the independent parameters and match the circuit length.~~

EDIT: In the end I decided to go for a simpler and cleaner solution, inspired by ``pennylane``'s approach to this matter.
Namely, I'm allowing the ``QuantumModel`` object to accept even `Callable` functions that construct a circuit given some input parameters (it is in fact very similar to an encoder, with the sole difference that the inputs are parameters of the model themselves). Therefore, to construct equivariant circuits, the user needs to define a function of the independent parameters that builds the circuit in object, e.g.:
```python
    # this defines 3 independent parameters
    def custom_circuit(engine, th, phi, lam):
        c = Circuit(2)
        delta = 2 * engine.cos(phi) + lam**2
        gamma = lam * engine.exp(th / 2)
        c.add([gates.RZ(i, theta=th) for i in range(2)])
        c.add([gates.RX(i, theta=lam) for i in range(2)])
        c.add([gates.RY(i, theta=phi) for i in range(2)])
        c.add(gates.RZ(0, theta=delta))
        c.add(gates.RX(1, theta=gamma))
        return c
```
Note that an additional `engine` argument is needed which will cover the role of `numpy`/`torch`/`tensorflow` or any engine to perform operations on the parameters. This is needed because at the moment we defer the circuit construction to the moment we calculate the custom gradients (with our `qiboml.differentiation.jax` and `qiboml.differentiation.PSR`) which then requires an engine compatible with the backend to be used  inside the custom function. 
This need could be lifted by directly constructing the whole circuit in the `interface`'s API (thus `torch` or `keras`), and delegating the calculation of the derivatives of any possible operations performed on the independent parameters to the interface as well, whereas the actual derivative wrt the circuit angles will be taken care of by our custom differentiation engine and correctly combined following the chain rule.  This is probably the best solution, as in order to make PSR work with equivariant models we would need to set up a similar structure anyway (thus at the moment PSR with equivariant models is not supported in this PR). However, this approach may involve a substantial refactor of our whole custom differentiation pipeline, which is maybe better to keep separated from this.
As a final note, keep in mind that each independent parameter has to be defined separately in the function and it's always inferred to be a scalar, since I use `inspect.signature` to retrieve the number of independent parameters and there would be no way (I believe) to infer the correct shape of each parameter from that.

That said, then the `QuantumModel` can be constructed as usual:
```python
model = QuantumModel([circuit, encoding, custom_circuit], decoding)
```

This should help with #108 and #99.

EDIT 2: with the new refactor #140 merged here the need for the engine to be passed to the function has been lifted, the appropriate engine for the interface should be used now. Thus, with ``torch``:
```python
    # this defines 3 independent parameters
    def custom_circuit(th, phi, lam):
        c = Circuit(2)
        delta = 2 * torch.cos(phi) + lam**2
        gamma = lam * torch.exp(th / 2)
        c.add([gates.RZ(i, theta=th) for i in range(2)])
        c.add([gates.RX(i, theta=lam) for i in range(2)])
        c.add([gates.RY(i, theta=phi) for i in range(2)])
        c.add(gates.RZ(0, theta=delta))
        c.add(gates.RX(1, theta=gamma))
        return c
```
whereas with ``keras``:
```python
    # this defines 3 independent parameters
    def custom_circuit(th, phi, lam):
        c = Circuit(2)
        delta = 2 * keras.ops.cos(phi) + lam**2
        gamma = lam * keras.ops.exp(th / 2)
        c.add([gates.RZ(i, theta=th) for i in range(2)])
        c.add([gates.RX(i, theta=lam) for i in range(2)])
        c.add([gates.RY(i, theta=phi) for i in range(2)])
        c.add(gates.RZ(0, theta=delta))
        c.add(gates.RX(1, theta=gamma))
        return c
```


MatteoRobbiati (commented):
Thanks a lot for this! A few comments follow, but we can go ahead.

PR #114, opened 2025-06-18
Add Variational Quantum Linear Solver (VQLS) Tutorial

ghaithalramahi:
Adding a new tutorial notebook implementing the Variational Quantum Linear Solver (VQLS) algorithm, based on the method introduced by Bravo-Prieto et al. (arXiv:1909.05820). This tutorial helps demonstrate how to use the QuantumModel and QuantumDecoding classes from QiboML to optimize circuit parameters.


src/qiboml/models/decoding.py
MatteoRobbiati, line 1: It is important that the decoders are agnostic to the operations we do, otherwise the machine learning frameworks will not be able to trace the operations and we will loose the gradients :)
For example, the current implementation relies on Numpy and Torch, and this will lead to problems when the decoder is used with the Keras interface. 
In Qiboml, every encoder has a `backend` attribute. It is actually a Qibo backend and what you can do is to rewrite the operations using `self.backend.np`. This constructs a common interface, but - under the hood - every backend is implementing the operations using a different engine (for example, `TensorFlowBackend.np` is instead using [`tensorflow.experimental.numpy`](https://www.tensorflow.org/api_docs/python/tf/experimental/numpy)).

src/qiboml/models/decoding.py
```diff
@@ -283,3 +286,38 @@ def output_shape(self) -> tuple[int, int]:
     @property
     def analytic(self) -> bool:  # pragma: no cover
         return False
+    
+@dataclass
+class VQLS(QuantumDecoding):
+    """Decoder for the Variational Quantum Linear Solver (VQLS).
```
MatteoRobbiati, line 292: You could add a reference to Carlos' paper here. Or explaining a bit more about the utility of this decoder :) 
This could also help the users to understand the notation of the arguments (why you have a target state, why there is a matrix `A`, etc).

src/qiboml/models/decoding.py
```diff
@@ -283,3 +286,38 @@ def output_shape(self) -> tuple[int, int]:
...
+        self.A = torch.tensor(self.A, dtype=torch.complex128)
+
+    def __call__(self, circuit: Circuit) -> torch.Tensor:
+        result = super().__call__(circuit)
+        state = result.state()  
+        final_state = self.A @ state
+        normalized = final_state / torch.linalg.norm(final_state)
+        overlap = torch.vdot(self.target_state, normalized)
+        return 1 - overlap.abs() ** 2
+
```
MatteoRobbiati, line 313: Please double check the suggestion here. In particular, I am not sure whether we should use here the `scipy.linalg` module instead of the `self.backend.np.linalg` one.

```suggestion
        super().__post_init__()
        self.target_state = self.backend.cast(self.target_state, dtype=self.backend.np.complex128)
        self.A = self.cast(self.A, dtype=self.backend.np.complex128)

    def __call__(self, circuit: Circuit) -> torch.Tensor:
        result = super().__call__(circuit)
        state = result.state()  
        final_state = self.A @ state
        normalized = final_state / self.backend.np.linalg.norm(final_state)
        overlap = self.backend.np.vdot(self.target_state, normalized)
        return 1 - overlap.abs() ** 2

```

tutorials/VQLS_tutorial.ipynb
MatteoRobbiati, line 1: You have probably updated a broken version of the notebook :)
Please check the errors.

MatteoRobbiati (changes requested):
Many thanks for this contribution!!
I left some comments below :)

src/qiboml/models/decoding.py
renatomello: To add, `backend.np.linalg.norm` can be replaced with the simpler `backend.calculate_vector_norm` (or `backend.calculate_matrix_norm` when required).

src/qiboml/models/decoding.py
```diff
@@ -308,7 +308,11 @@ def __post_init__(self):
         
     def __call__(self, circuit: Circuit):
         result = super().__call__(circuit)
-        state = result.state()  
+        state = result.state()
+
+        if self.A is None or self.target_state is None:
+            raise ValueError("Both A and target_state must be provided.")
```
renatomello, line 314:
```suggestion
            raise ValueError("Both A and target_state must be provided.")  
```
```suggestion
            raise_error(ValueError, "Both A and target_state must be provided.")  
```

src/qiboml/models/decoding.py
```diff
@@ -283,3 +285,47 @@ def output_shape(self) -> tuple[int, int]:
...
+    
+    Ensure QiboML backend set to Pytorch. 
+    """
+    target_state: ndarray = None
+    A: ndarray = None
+
+    def __post_init__(self):
+        super().__post_init__()
+        self.target_state = self.backend.cast(self.target_state, dtype=self.backend.np.complex128)
+        self.A = self.backend.cast(self.A, dtype=self.backend.np.complex128)
```
renatomello, line 306: Do they necessarily need to be `complex`?

src/qiboml/models/decoding.py
```diff
@@ -283,3 +285,47 @@ def output_shape(self) -> tuple[int, int]:
...
+        result = super().__call__(circuit)
+        state = result.state()
+
+        if self.A is None or self.target_state is None:
+            raise_error(ValueError, "Both ``A`` and ``target_state`` must be provided.")  
+        
+        final_state = self.A @ state
+        normalized = final_state / self.backend.calculate_vector_norm(final_state)
+        overlap = self.backend.np.vdot(self.target_state, normalized)
+        cost = 1 - overlap.abs() ** 2
```
renatomello, line 319:
```suggestion
        cost = infidelity(normalized, target, backend=self.backend)
```

While importing at top level `from qibo.quantum_info.metrics import infidelity`.

src/qiboml/models/decoding.py
```diff
@@ -283,3 +286,46 @@ def output_shape(self) -> tuple[int, int]:
...
+    Adapted from the following paper arXiv:1909.05820v4 by Carlos Bravo-Prieto et al. 
+
+    Args:
+        target_state (ndarray): Target solution vector :math:`\\ket{b}`.
+        A (ndarray): The matrix ``A`` in the linear system :math:`A \\, \\ket{x} = \\ket{b}`.
+    
+    Ensure QiboML backend set to Pytorch. 
+    """
+    target_state: ndarray = None
+    A: ndarray = None
```
BrunoLiegiBastonLiegi, line 302:
```suggestion
    target_state: ndarray
    A: ndarray
```
are these actually optional? because you raise the error below when they are missing

BrunoLiegiBastonLiegi (commented):
Thanks, I have just a comment. If tests are passing, this is ready to be merged in my opinion.

renatomello:
@ghaithalramahi can you check why tests are failing?

MatteoRobbiati (commented):
Some tests are failing because, in `test_models_interfaces` we sample random decoders and execute. When your decoder is sampled, there are no `A` and `target_state` passed. 
I would follow one of these two options: 

1. write a dedicated test for this decoder and remove this one from the list of decoders sampled in the aforementioned test.
2. when your decoder is sampled, we define matrix and target state and pass it to the decoder initialization.

src/qiboml/models/decoding.py
```diff
@@ -378,6 +379,47 @@ def output_shape(self) -> tuple[int, int]:
...
+@dataclass(kw_only=True)
+class VariationalQuantumLinearSolver(QuantumDecoding):
+    """Decoder for the Variational Quantum Linear Solver (VQLS).
+    Adapted from the following paper arXiv:1909.05820v4 by Carlos Bravo-Prieto et al. 
+
+    Args:
+        target_state (ndarray): Target solution vector :math:`\\ket{b}`.
+        A (ndarray): The matrix ``A`` in the linear system :math:`A \\, \\ket{x} = \\ket{b}`.
+    
+    Ensure QiboML backend set to Pytorch.
```
renatomello, line 392:
```suggestion
    Adapted from the following paper arXiv:1909.05820v4 by Carlos Bravo-Prieto et al. 

    .. note::
        Currently, only implemented for the ``pytorch`` backend.

    Args:
        target_state (ndarray): Target solution vector :math:`\\ket{b}`.
        A (ndarray): The matrix ``A`` in the linear system :math:`A \\, \\ket{x} = \\ket{b}`.
```

renatomello: Also proper article referencing like in #122

MatteoRobbiati: Yes thanks! 
Think that, here in Qiboml, every decoder is supposed to be used within an optimization pipeline. Do not worry about it, because we have mechanisms to retrieve the gradients implemented in the package. If you are interested, you can have a look into `qiboml.interfaces` :)

renatomello:
@ghaithalramahi @BrunoLiegiBastonLiegi @MatteoRobbiati what's the status here?

BrunoLiegiBastonLiegi (approved):

PR #121, opened 2025-07-04
Add tutorial on real time error mitigation to documentation

MatteoRobbiati:
As per title.


Edoardo-Pedicillo (approved):

MatteoRobbiati:
> Thanks, nice one. Just one thing, do we actually need to pass `density_matrix=True` do the `HardwareEfficient` object on cell 59? I think we could drop that.

Thanks! I realized there was a problem with the mitigation part, because I was building the circuit before  aligning the `density_matrix` attribute.
Now it works.

PR #124, opened 2025-07-07
Calibrator implemetation

Edoardo-Pedicillo:


tests/test_calibrator.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+    optimizer = torch.optim.Adam(model.parameters(), lr=0.1)
+
+    for epoch in range(epochs):
+        optimizer.zero_grad()
+        cost = model()
+        cost.backward()
+        optimizer.step()
+
+    assert dec.calibrator._history != None
```
BrunoLiegiBastonLiegi, line 105: maybe also check that the history makes sense, e.g. it contains what you expect it to contain

src/qiboml/models/utils.py
```diff
@@ -142,6 +151,58 @@ def data_regression(
...
+
+    def __call__(self):
+        self._counter += 1
+        if self._counter % self.trigger_shots == 0:
+            self.execute_experiments()
+
+    def execute_experiments(self):
+        """Execute the experiments in the runcard."""
+        platform = self.backend.platform
+        assert platform is not None, "Invalid None platform"
```
andrea-pasquale, line 180:
```suggestion
        assert platform is not None
```
I think that the error message is redundant in this case.

src/qiboml/models/utils.py
```diff
@@ -142,6 +151,58 @@ def data_regression(
...
+class Calibrator:
+
+    runcard: Runcard
+    """Qibocal runcard with the calibration protocol to be executed."""
+    path: Path
+    """Folder to dump the protocol's data and results."""
+    backend: Backend
+    """Qibo backend."""
+    trigger_shots: int = 100
+    """NUmber of shots to trigger :meth:`qiboml.models.utils.Calibrator.execute_experiments`"""
```
andrea-pasquale, line 164:
```suggestion
    """Number of shots to trigger :meth:`qiboml.models.utils.Calibrator.execute_experiments`"""
```

andrea-pasquale (commented):
I think that the current implementation is mostly fine. Some refactoring is necessary to avoid importing qibocal if we want to have it as an optional dep.

src/qiboml/models/utils.py
```diff
@@ -142,6 +151,58 @@ def data_regression(
...
+            )
+        )
+
+    def data(self, id: str, execution_number: int):
+        """Return the data of the protocol with the specific `id` at its
+        `execution_number` execution.
+        """
+        return self._history[execution_number][id][0].data
+
+    def results(self, id: str, execution_number: int):
```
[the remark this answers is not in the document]
Edoardo-Pedicillo: I prefer everything to be explicit.

src/qiboml/models/calibrator.py
```diff
@@ -0,0 +1,67 @@
...
+    @dataclass
+    class Calibrator:
+
+        runcard: Runcard
+        """Qibocal runcard with the calibration protocol to be executed."""
+        path: Path
+        """Folder to dump the protocol's data and results."""
+        backend: Backend
+        """Qibo backend."""
+        trigger_shots: int = 100
```
MatteoRobbiati, line 24: Same comments of before about the name of the variable.

PR #125, opened 2025-07-08
Expectation from circuit

BrunoLiegiBastonLiegi:
Use qibo's expectation from circuit to allow for non diagonal observables.


src/qiboml/models/decoding.py
```diff
@@ -84,19 +85,30 @@ def __call__(
...
+        x = self.transpile(x)
+        x = self.apply_noise(x)
+        return x
+
+    def align_circuits(self, x: Circuit):
         # Standardize the density matrix attribute
         self._align_density_matrix(x)
         self._align_wire_names(x)
 
+    def transpile(self, x: Circuit) -> Circuit:
```
MatteoRobbiati, line 103:
```suggestion
    def transpile(self, x: Circuit) -> Circuit:
        """Transpile a given circuit ``x`` using the instructions provided by the ``transpiler`` attribute."""
```

MatteoRobbiati: I am adding a small docstrings since the method is defined public.

src/qiboml/models/decoding.py
```diff
@@ -84,19 +85,30 @@ def __call__(
...
         if self.transpiler is not None:
             x, _ = self.transpiler(x)
+        return x
 
+    def apply_noise(self, x: Circuit) -> Circuit:
```
MatteoRobbiati, line 108:
```suggestion
    def apply_noise(self, x: Circuit) -> Circuit:
        """Apply the decoder ``noise_model`` to the target circuit."""
```

src/qiboml/models/decoding.py
```diff
@@ -84,19 +85,30 @@ def __call__(
         Returns:
             (CircuitResult | QuantumState | MeasurementOutcomes): the execution ``qibo.result`` object.
         """
+        x = self.preprocessing(x)
+
+        return self.backend.execute_circuit(x + self._circuit, nshots=self.nshots)
+
+    def preprocessing(self, x: Circuit) -> Circuit:
```
MatteoRobbiati, line 92:
```suggestion
    def _preprocessing(self, x: Circuit) -> Circuit:
```

If we don't plan to expose it.
If you agree with this, we also have to update the name everywhere it is used.

MatteoRobbiati (commented):
Thanks! I left just a few comments.

src/qiboml/models/decoding.py
MatteoRobbiati: OK, fine! Then I think we just need to add a oneline docstring here as well :)

MatteoRobbiati (approved):

src/qiboml/models/decoding.py
```diff
@@ -374,7 +396,7 @@ def __call__(self, x: Circuit) -> ndarray:
         Returns:
             (ndarray): the generated samples.
```
renatomello, line 397:
```suggestion
            ndarray: Generated samples.
```

renatomello (approved):

PR #132, opened 2025-07-17
Adding extras

andrea-pasquale:
Just a quick fixes adding extras for `torch`, `jax` and `tensorflow` which are already assumed in the documentation.


andrea-pasquale:
I've noticed that we need to keep `jax` as a dependency in order to have the jitted functions here

https://github.com/qiboteam/qiboml/blob/47a59b680680a145e077bc6e346f612cdb136eb2/src/qiboml/backends/jax.py#L12-L58

At this point can't we ship qiboml with all dependencies (including both `tensorflow` and `PyTorch`)?
Otherwise we can just have those as optional dependencies and keep jax as the only required dependency.

BrunoLiegiBastonLiegi (approved):

PR #133, opened 2025-07-18
``HardwareEfficient`` ansatz fix

BrunoLiegiBastonLiegi:
As per title.


BrunoLiegiBastonLiegi:
There is a test failing for reasons that are obscure to me. I will have to look into that one first.

PR #136, opened 2025-07-28
Include more circuit ansatze in `models.ansatze`

renatomello:


BrunoLiegiBastonLiegi (commented):
Thanks! this looks ready to go to me. Have you checked that this is correctly (automatically) tested in `test_interfaces.py::test_encoding`? it should be since the coverage is fine, but better to doublecheck anyway.

PR #146, opened 2025-08-24
Update observable's backend when computing reference value in mitigation procedures

MatteoRobbiati:
As per title.

### DIFF
--- a/src/qiboml/models/utils.py
+++ b/src/qiboml/models/utils.py
@@ -75,6 +75,14 @@ class Mitigator:
         circuit: Circuit,
     ):
         """Construct reference error sensitive circuit."""
+        # Ensuring the observable backend is the simulation one
+        matrix = self._simulation_backend.to_numpy(observable.matrix)
+        observable = Hamiltonian(
+            nqubits=circuit.nqubits,
+            matrix=self._simulation_backend.cast(matrix),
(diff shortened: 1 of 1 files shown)


src/qiboml/models/utils.py
```diff
@@ -75,6 +75,14 @@ def calculate_reference_expval(
...
     ):
         """Construct reference error sensitive circuit."""
+        # Ensuring the observable backend is the simulation one
+        matrix = self._simulation_backend.to_numpy(observable.matrix)
+        observable = Hamiltonian(
+            nqubits=circuit.nqubits,
+            matrix=self._simulation_backend.cast(matrix),
+            backend=self._simulation_backend,
+        )
+
```
BrunoLiegiBastonLiegi, line 85: You should probably check whether it's a Symbolic or Dense hamiltonian, and act differently for each one.
```suggestion
        if isinstance(observable, SymbolicHamiltonian):
            observable = SymbolicHamiltonian(observable.form, nqubits=observable.nqubits, backend=self._simulation_backend)
        else:
            matrix = observable.backend.to_numpy(observable.matrix)
            observable = Hamiltonian(
                nqubits=circuit.nqubits,
                matrix=self._simulation_backend.cast(matrix),
                backend=self._simulation_backend,
            )

```

PR #151, opened 2025-09-23
Hardware efficient ansatz

AlejandroSopena:
This PR generalizes the HEA allowing custom single-qubit and entangling blocks.


src/qiboml/models/ansatze.py
```diff
@@ -12,21 +12,74 @@ def HardwareEfficient(
...
+    closed_boundary: bool = True,
+    **kwargs,
 ) -> Circuit:
+    """
+    Create a Hardware-efficient ansatz with custom single-qubit and entangling blocks.
+
+    Args:
+        nqubits (int): Number of qubits in the ansatz.
+        qubits (tuple[int], optional): Qubit indices to apply the ansatz to. If ``None``, the ansatz is applied to all qubits
+            from ``0`` to ``nqubits-1``. Defaults to ``None``.
```
renatomello, line 28:
```suggestion
        qubits (tuple[int], optional): Qubit indexes to apply the ansatz to. If ``None``, 
            the ansatz is applied to all qubits from :math:`0` to :math:`nqubits-1`. 
            Defaults to ``None``.
```

src/qiboml/models/ansatze.py
```diff
@@ -12,21 +12,74 @@ def HardwareEfficient(
...
+        nlayers (int, optional): Number of layers (single-qubit + entangling per layer). Defaults to 1.
```
renatomello, line 29:
```suggestion
        nlayers (int, optional): Number of layers (single-qubit + entangling per layer). Defaults to :math:`1`.
```

src/qiboml/models/ansatze.py
```diff
@@ -12,21 +12,74 @@ def HardwareEfficient(
...
+        single_block (Circuit, optional): 1-qubit Circuit applied to each qubit. Defaults to a block with :class:`qibo.gates.RY` and :class:`qibo.gates.RZ` gates.
+        entangling_block (Circuit, optional): full n-qubit entangling circuit. Defaults to ``None``.
```
renatomello, line 31:
```suggestion
        entangling_block (Circuit, optional): :math:`n`-qubit entangling circuit. Defaults to ``None``.
```
What happens if `None`?

src/qiboml/models/ansatze.py
```diff
@@ -12,21 +12,74 @@ def HardwareEfficient(
...
+        entangling_gate (str or :class:`qibo.gates.Gate`, optional): Only used if ``entangling_block`` is None. Two-qubit gate to be used
+            in the entangling layer if ``entangling_block`` is not provided. If ``entangling_gate`` is a parametrized gate,
+            all phases are initialized as :math:`0.0`. Defaults to  ``"CNOT"``.
```
renatomello, line 34:
```suggestion
        entangling_gate (str or :class:`qibo.gates.Gate`, optional): Only used if ``entangling_block``
        is ``None``. Two-qubit gate to be used in the entangling layer if ``entangling_block`` is not
        provided. If ``entangling_gate`` is a parametrized gate, all phases are initialized as 
        :math:`0.0`. Defaults to  ``"CNOT"``.
```

src/qiboml/models/ansatze.py
```diff
@@ -12,21 +12,74 @@ def HardwareEfficient(
...
+        architecture (str, optional): Only used if ``entangling_block`` is None. Architecture of the entangling layer.
+            In alphabetical order, options are ``"diagonal"``, ``"even_layer"``,
+            ``"next_nearest"``, ``"odd_layer"``, ``"pyramid"``, ``"shifted"``,
+            ``"v"``, and ``"x"``. The ``"x"`` architecture is only defined for an even number
+            of qubits. Defaults to ``"diagonal"``.
+        closed_boundary (bool, optional): Only used if ``entangling_block`` is None. If ``True`` and ``architecture not in
+            ["pyramid", "v", "x"]``, adds a closed-boundary condition to the entangling layer.
+            Defaults to ``True``.
```
renatomello, line 42:
```suggestion
        closed_boundary (bool, optional): Only used if ``entangling_block`` is ``None``.
            If ``True`` and ``architecture not in ["pyramid", "v", "x"]``, adds a 
            closed-boundary condition to the entangling layer. Defaults to ``True``.
```

src/qiboml/models/ansatze.py
```diff
@@ -12,21 +12,74 @@ def HardwareEfficient(
...
+    """
+    circ = Circuit(nqubits, **kwargs)
+
     if qubits is None:
         qubits = list(range(nqubits))
-    circuit = Circuit(nqubits)
+    elif len(qubits) > nqubits:
+        raise ValueError(
+            f"Number of specified qubits ({len(qubits)}) cannot exceed the total number of qubits in the circuit ({nqubits})."
+        )
```
renatomello, line 56:
```suggestion
        raise_error(
            ValueError,
            f"Number of specified qubits ({len(qubits)}) cannot exceed the total number of qubits in the circuit ({nqubits})."
        )
```

BrunoLiegiBastonLiegi (commented):
Thanks, all good, just a small comment.

src/qiboml/models/ansatze.py
```diff
@@ -4,29 +4,86 @@
...
+    elif len(qubits) > nqubits:
+        raise_error(
+            ValueError,
+            f"Number of specified qubits ({len(qubits)}) cannot exceed the total number of qubits in the circuit ({nqubits})."
+        )
+
+    if single_block is None:
+        single_block = Circuit(1)
+        single_block.add(gates.RY(0, theta=0.0, trainable=True))
+        single_block.add(gates.RZ(0, theta=0.0, trainable=True))
```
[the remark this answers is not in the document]
renatomello: It is very weird to me that a circuit is initialized with non-zero amplitudes. Even if the goal of the user is, in the end, to set random phases, that's a user choice (as well as the distribution to sample from). Right now, not only are phases being imposed, but also their distribution.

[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: It's the standard practice in ML to initialize the parameters of the layers randomly under some distribution of choice. Not sure what's the best practice in QML, but in my opinion, it's better to start from a random configuration that changes every time (unless the seed is fixed), not necessarily the one we use right now.
In literature several methods have been proposed for parameter initialization apparently:
- [https://quantum-journal.org/papers/q-2019-12-09-214/](https://quantum-journal.org/papers/q-2019-12-09-214/)
- [https://arxiv.org/abs/2302.06858](https://arxiv.org/abs/2302.06858)
- [https://arxiv.org/abs/2112.10952v1](https://arxiv.org/abs/2112.10952v1)
- [https://ieeexplore.ieee.org/document/10546644](https://ieeexplore.ieee.org/document/10546644)
- [https://arxiv.org/abs/2207.01539](https://arxiv.org/abs/2207.01539)
- [https://arxiv.org/abs/2103.08572](https://arxiv.org/abs/2103.08572)

we can just pick one, the most simple for now.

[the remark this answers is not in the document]
renatomello: I'm well aware that this is standard practice. The point here is when these random phases are set. I don't think it makes sense to do it embedded with the circuit generation.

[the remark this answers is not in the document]
BrunoLiegiBastonLiegi: Why not? Then in the tests don't set the angles manually to random values, leave them to zero, otherwise we're just making this more clunky for the user. It all boils down to what an user will most often want to run a trainable layer with, out of the box.

PR #221, opened 2026-06-30
Qcnn devel

ihpcganck:
qcnn.py was copied from qibo/models but it should be put under qiboml/models. Now it supports 3 variants of QCNN, i.e., QCNNCOMPLEX, QCNNREAL, QCNNRY.

### DIFF
--- a/src/qiboml/models/qcnn.py
+++ b/src/qiboml/models/qcnn.py
@@ -88,9 +88,6 @@ class QuantumCNN:
             ValueError: If nqubits is not larger than 1.
         """
 
-
-        # assert False, 'stop here!'
-
         self.quantum_model = quantum_model
         self.nqubits = nqubits
         if self.nqubits <= 1:
@@ -99,8 +96,6 @@ class QuantumCNN:
         self.nclasses = nclasses
         self.nlayers = nlayers
 
-
-        twoqubitansatz = None
         if quantum_model == 'QCNNRY': # special handling
           twoqubitansatz = Circuit(2)
           twoqubitansatz.add( gates.RY(0,0))
@@ -340,7 +335,7 @@ class QuantumCNN:
 
         self._circuit.set_parameters(expanded_params)
 
-    def Classifier_circuit(self, theta):
+    def classifier_circuit(self, theta):
         """
         Args:
             theta: list or numpy.array with the biases and the angles to be used in the circuit.
@@ -353,7 +348,7 @@ class QuantumCNN:
         self.set_circuit_params(angles)
         return self._circuit
 
-    def Predictions(self, circuit, theta, init_state, nshots=10000):
+    def circuit_predictions(self, circuit, theta, init_state, nshots=10000):
         """
         Args:
             theta: list or numpy.array with the biases to be used in the circuit.
@@ -394,7 +389,7 @@ class QuantumCNN:
 
         return loss / len(labels)
 
-    def Cost_function(self, theta, data=None, labels=None, nshots=10000):
+    def cost_function(self, theta, data=None, labels=None, nshots=10000):
         """
         Args:
             theta: list or numpy.array with the biases and the angles to be used in the circuit.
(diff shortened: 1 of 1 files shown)


src/qiboml/models/qcnn.py
```diff
@@ -0,0 +1,478 @@
...
+            twoqubitansatz (qibo.models.circuit.Circuit, optional): A two-qubit ansatz for the convolutional layers. If None, a default ansatz is used. Defaults to None.
+            copy_init_state (bool, optional): Whether to copy the initial state for each shot in the simulation. If None, the behavior depends on the backend. Defaults to None.
+            fullstate: Default to None. True means we ignore the measurement gates
+
+        Raises:
+            ValueError: If nqubits is not larger than 1.
+        """
+
+
+        # assert False, 'stop here!'
```
shangtai: I believe line 92 is a comment left during debugging, remove it.

src/qiboml/models/qcnn.py
```diff
@@ -0,0 +1,478 @@
...
+        self.quantum_model = quantum_model
+        self.nqubits = nqubits
+        if self.nqubits <= 1:
+            raise ValueError("nqubits must be larger than 1")
+
+        self.nclasses = nclasses
+        self.nlayers = nlayers
+
+
+        twoqubitansatz = None
```
shangtai: twoqubitansatz is an input to the init in line 70 right? If we set it to None here, it would defeat the purpose of taking in the input right?

src/qiboml/models/qcnn.py
```diff
@@ -0,0 +1,478 @@
...
+            pool_params = params[
+                param_start + self.nparams_conv : param_start + self.nparams_layer
+            ]
+            pool_params += [-pool_params[2], -pool_params[1], -pool_params[0]]
+            expanded_params += conv_params * int(nleft if nleft > 2 else 1)
+            expanded_params += pool_params * int(nleft / 2)
+
+        self._circuit.set_parameters(expanded_params)
+
+    def Classifier_circuit(self, theta):
```
shangtai: Consider renaming it to classifier_circuit, i.e. using snake_case.

src/qiboml/models/qcnn.py
```diff
@@ -0,0 +1,478 @@
...
+        Returns:
+            Circuit implementing the variational ansatz for angles "theta".
+        """
+        bias = np.array(theta[0 : self.measured_qubits])
+        angles = theta[self.measured_qubits :]
+
+        self.set_circuit_params(angles)
+        return self._circuit
+
+    def Predictions(self, circuit, theta, init_state, nshots=10000):
```
shangtai: Usually we use snake_case for method name.

src/qiboml/models/qcnn.py
```diff
@@ -0,0 +1,478 @@
...
+            numpy.float32 with the value of the square-loss function.
+        """
+        loss = 0
+        for l, p in zip(labels, predictions):
+            for qubit in range(self.measured_qubits):
+                loss += (l[qubit] - p[qubit]) ** 2
+
+        return loss / len(labels)
+
+    def Cost_function(self, theta, data=None, labels=None, nshots=10000):
```
shangtai: same as above, consider changing it to cost_function.

src/qiboml/models/qcnn.py
```diff
@@ -0,0 +1,478 @@
...
+
+        return circuit
+
+    def two_qubit_unitary(self, bits, symbols):
+        """
+        Internal helper function to create a circuit consisting of two qubit unitaries.
+
+        Args:
+            bits: the two qubits to apply the unitaries to
+            symbols: length 15 array containing the parameters
```
shangtai, line 263: bits is assumed to be of length 15, would be good to use assertion to catch if a user input bits of different lengths.

src/qiboml/models/qcnn.py
```diff
@@ -0,0 +1,478 @@
...
+        return circuit
+
+    def one_qubit_unitary(self, bit, symbols):
+        """
+        Internal helper function to make a circuit enacting a rotation of the bloch sphere about the X,
+        Y and Z axis, that depends on the values in `symbols`.
+
+        Args:
+            bit: the qubit to apply the one-qubit unitaries to
+            symbols: length 3 array containing the parameters
```
shangtai, line 246: use assertion to validate that symbols have length 3.

shangtai (changes requested):
Reminder to include pytests and update the documentations as well.

scarrazza:
@shangtai could you please have a look at these changes. @renatomello @MatteoRobbiati could you please check if this implementation looks consistent with QiboML?

src/qiboml/models/qcnn.py
shangtai: Assertions not done yet, is this by choice?

src/qiboml/models/qcnn.py
shangtai: same as above, no check of length.

shangtai (commented):
Are we including any pytest for Qcnn?

